PR #69, opened 2023-09-20
Webserver and yagna autostart

approxit:
What I've done:
- Upgraded yagna autostart/stop in webserver - yagna is checked / started at webserver start, and stopped at webserver stop (if yagna was managed by webserver in the first place)
- Enabled auto `yagna payment fund`, as its required each time after yagna restart, due to usage of goerli instead of rinkeby
- Upgraded a little logs of few things to have them more explicit
- Upgraded webserver autostart/stop in node provider - webserver is checked at before any webserver call, started if needed, and stopped after all nodes are terminated
- Reformatted the code

Notable remarks:
- Auto `yagna payment fund` should be tested / changed while using mainnet, as the is no reason to force the user to wait to collect testnet funds
- As node provider is used by ray only for atomic cluster actions, we have no control over `up` and `down` commands, hence we don't know the true intent. This arises in webserver stopping, node termination is not a last commands that is ran at `down`. I've assumed that 10s delay from

### DIFF
--- a/ray_on_golem/client/client.py
+++ b/ray_on_golem/client/client.py
@@ -1,14 +1,20 @@
+# TODO: Consider moving this package to ray_on_golem/provider/client because of cli_logger usage
+import subprocess
 from functools import lru_cache
 from http import HTTPStatus
+from time import sleep
 from typing import Any, Dict, List, Type, TypeVar
 
 import requests
 from pydantic import BaseModel, ValidationError
+from ray.autoscaler._private.cli_logger import cli_logger
 from yarl import URL
 
 from ray_on_golem.client.exceptions import RayOnGolemClientError, RayOnGolemClientValidationError
 from ray_on_golem.server import models, settings
-from ray_on_golem.server.models import NodeConfigData
+from ray_on_golem.server.models import NodeConfigData, ShutdownState
+from ray_on_golem.server.settings import RAY_ON_GOLEM_PATH, URL_HEALTH_CHECK
+from ray_on_golem.utils import is_running_on_golem_network
 
 TResponseModel = TypeVar("TResponseModel")
 
@@ -18,6 +24,9 @@ class RayOnGolemClient:
         self._base_url = base_url
 
         self._session = requests.Session()
+
+        if not is_running_on_golem_network():
+            self._start_webserver()
 
     @classmethod
     @lru_cache()
@@ -73,6 +82,8 @@ class RayOnGolemClient:
             response_model=models.TerminateNodeResponseData,
             error_message="Couldn't terminate node",
         )
+
+        self._stop_webserver()
 
         return response.terminated_nodes
 
@@ -192,3 +203,62 @@ class RayOnGolemClient:
             raise RayOnGolemClientValidationError(
                 "Couldn't validate response data",
             ) from e
+
+    def _start_webserver(self) -> None:
+        with cli_logger.group("Ray On Golem webserver"):
+            if self._is_webserver_running():
+                cli_logger.print("Webserver is already running")
+                return
+
+            cli_logger.print("Starting webserver...")
+            subprocess.Popen(
+                [RAY_ON_GOLEM_PATH, "-p", str(self._base_url.port), "--self-shutdown"],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                start_new_session=True,
+            )
+
+            for _ in range(10):
+                sleep(2)
+
+                if self._is_webserver_running():
+                    cli_logger.print("Starting webserver done")
+                    return
+
+                cli_logger.print("Webserver is not yet running, waiting additional 2 seconds...")
+
+            cli_logger.abort("Starting webserver failed!")
+
+    def _stop_webserver(self) -> None:
+        with cli_logger.group("Ray On Golem webserver"):
+            if not self._is_webserver_running():
+                cli_logger.print("Webserver is already stopped")
+                return
+
+            cli_logger.print("Requesting webserver stop...")
+
+            response = self._make_request(
+                url=settings.URL_SELF_SHUTDOWN,
+                request_data=models.SelfShutdownRequestData(),
+                response_model=models.SelfShutdownResponseData,
+                error_message="Couldn't send self shutdown request",
+            )
+
+            if response.shutdown_state == ShutdownState.NOT_ENABLED:
+                cli_logger.print("No need to stop webserver, as it was ran externally")
+                return
+            elif response.shutdown_state == ShutdownState.CLUSTER_NOT_EMPTY:
+                cli_logger.print("No need to stop webserver, as cluster is not empty")
+                return
+
+            cli_logger.print("Requesting webserver done, will stop soon")
+
+    def _is_webserver_running(self) -> bool:
+        try:
+            response = self._session.get(
+                str(self._base_url / URL_HEALTH_CHECK.lstrip("/")), timeout=2
+            )
+        except requests.ConnectionError:
+            return False
+        else:
+            return response.status_code == HTTPStatus.OK and response.text == "ok"
--- a/ray_on_golem/server/models.py
+++ b/ray_on_golem/server/models.py
@@ -14,6 +14,12 @@ class NodeState(Enum):
     stopping = "stopping"
 
 
+class ShutdownState(Enum):
+    NOT_ENABLED = "not_enabled"
+    CLUSTER_NOT_EMPTY = "cluster_not_empty"
+    WILL_SHUTDOWN = "will_shutdown"
+
+
 class Node(BaseModel):
     node_id: NodeId
     state: NodeState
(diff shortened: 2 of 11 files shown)


ray_on_golem/server/services/yagna.py
```diff
@@ -26,23 +30,18 @@ def __init__(self, yagna_path: Path):
         self._yagna_process: Optional[Process] = None
 
     async def init(self) -> None:
-        self.yagna_appkey = await self._get_or_create_yagna_appkey()
-
         if await self._check_if_yagna_is_running():
             logger.info("Yagna service is already running")
         else:
             await self._run_yagna_service()
+            await self._run_yagna_payment_fund()  # FIXME
```
shadeofblue, line 37: `yagna payment fund` is needed only when yagna is launched the first time - or - when it depletes testnet fund (be it ETH or tGLM)

on a regular restart, only `yagna payment init` is currently required (and, actually, could be done without if we ported the mechanism currently used in yapapi)

ray_on_golem/client/client.py
```diff
@@ -19,6 +25,9 @@ def __init__(self, base_url: URL) -> None:
 
         self._session = requests.Session()
 
+        if not is_running_on_golem_network():
+            self._start_webserver()
```
shadeofblue, line 29: wondering aloud... is it sensible to put such long-blocking operations into a constructor? my intuition suggests this should be a separate step...

plus, it sounds somewhat unintuitive in general to start a _server_ when you instiantiate a _client_ ...

shadeofblue, line 29: on a separate note - question - what if the webserver startup fails when it's required? wouldn't we want to raise an explicit exception?

ray_on_golem/client/client.py
```diff
@@ -192,3 +203,62 @@ def _make_request(
...
+
+            cli_logger.print("Starting webserver...")
+            subprocess.Popen(
+                [RAY_ON_GOLEM_PATH, "-p", str(self._base_url.port), "--self-shutdown"],
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                start_new_session=True,
+            )
+
+            for _ in range(10):
```
shadeofblue, line 221: nitpicking but possibly such magic values as the number of retries and the interval between them should be extracted to module-level constants ...

ray_on_golem/client/client.py
```diff
@@ -192,3 +203,62 @@ def _make_request(
             raise RayOnGolemClientValidationError(
                 "Couldn't validate response data",
             ) from e
+
+    def _start_webserver(self) -> None:
+        with cli_logger.group("Ray On Golem webserver"):
```
shadeofblue, line 208: the name of the logging group should most likely be a constant defined somewhere...

ray_on_golem/client/client.py
```diff
@@ -192,3 +203,62 @@ def _make_request(
...
+                cli_logger.print("Webserver is already stopped")
+                return
+
+            cli_logger.print("Requesting webserver stop...")
+
+            response = self._make_request(
+                url=settings.URL_SELF_SHUTDOWN,
+                request_data=models.SelfShutdownRequestData(),
+                response_model=models.SelfShutdownResponseData,
+                error_message="Couldn't send self shutdown request",
```
shadeofblue, line 244:
```suggestion
                error_message="Couldn't send a self-shutdown request",
```

ray_on_golem/client/client.py
```diff
@@ -192,3 +203,62 @@ def _make_request(
...
+            )
+
+            if response.shutdown_state == ShutdownState.NOT_ENABLED:
+                cli_logger.print("No need to stop webserver, as it was ran externally")
```
shadeofblue, line 248:
```suggestion
                cli_logger.print("No need to stop the webserver, as it was started externally")
```

ray_on_golem/client/client.py
```diff
@@ -192,3 +203,62 @@ def _make_request(
...
+                return
+            elif response.shutdown_state == ShutdownState.CLUSTER_NOT_EMPTY:
+                cli_logger.print("No need to stop webserver, as cluster is not empty")
```
shadeofblue, line 251:
```suggestion
                cli_logger.print("No need to stop the webserver, as the cluster is not empty")
```

ray_on_golem/client/client.py
```diff
@@ -192,3 +203,62 @@ def _make_request(
...
+                return
+
+            cli_logger.print("Requesting webserver done, will stop soon")
```
shadeofblue, line 254: maybe this way?

```suggestion
            cli_logger.print("Webserver shutdown request done, will stop soon")
```

ray_on_golem/client/client.py
```diff
@@ -192,3 +203,62 @@ def _make_request(
...
+
+    def _is_webserver_running(self) -> bool:
+        try:
+            response = self._session.get(
+                str(self._base_url / URL_HEALTH_CHECK.lstrip("/")), timeout=2
```
shadeofblue, line 259: I'd define this timeout also as some constant on a module level... 

(or, we could have them as parameters provided to the class's constructor)

ray_on_golem/server/services/yagna.py
approxit: Changed to `payment init`.

ray_on_golem/client/client.py
approxit: Well... It sounded better at first, but good point. Moving this code to node provider fixes the problem of ray logs while starting webserver.

ray_on_golem/server/models.py
```diff
@@ -14,6 +14,12 @@ class NodeState(Enum):
     stopping = "stopping"
 
 
+class ShutdownState(Enum):
```
shadeofblue, line 17: maybe not now, but if, in the future, we want to support various transitions of the server/cluster state, this might benefit from being a state machine...

ray_on_golem/server/run.py
```diff
@@ -86,8 +111,10 @@ def main():
     args = parse_sys_args()
     prepare_tmp_dir()
 
-    app = create_application()
-    web.run_app(app, port=args.port)
+    app = create_application(args.port, args.self_shutdown)
+    web.run_app(app, port=args.port, print=None)
```
shadeofblue, line 115: hmm, if we're already using the `app`'s setitem/getitem interface to store arbitrary additional properties, we could store the `port` there as well ... and then, there would be just _one_ place where we define the server's port... 

```suggestion
    app = create_application(args.port, args.self_shutdown)
    web.run_app(app, port=app["port"], print=None)
```

plus, (not not, just a thought) ->  if we're allowing the `port` to be chosen, maybe we should also allow override of the `host` argument?

ray_on_golem/server/services/yagna.py
```diff
@@ -5,13 +5,17 @@
...
+from yarl import URL
+
 from ray_on_golem.exceptions import RayOnGolemError
 from ray_on_golem.server.settings import YAGNA_APPKEY
 from ray_on_golem.utils import run_subprocess
 
 logger = logging.getLogger(__name__)
 
 YAGNA_APPNAME = "ray-on-golem"
+YAGNA_API_URL = URL("http://127.0.0.1:7465")
```
shadeofblue, line 18: don't we want to allow this to be overridden? by an environment variable perhaps?

ray_on_golem/server/services/yagna.py
```diff
@@ -68,14 +67,37 @@ async def _run_yagna_payment_fund(self) -> bool:
...
         if is_running:
             self._yagna_process = process
+            logger.info("Starting Yagna service done")
         else:
-            logger.error("Failed to start yagna!")
+            logger.error("Starting Yagna service failed!")
+
+    async def _stop_yagna_service(self):
+        if self._yagna_process is None:
+            logger.info("No need to stop Yagna service, as it was ran externally")
```
shadeofblue, line 90:
```suggestion
            logger.info("No need to stop Yagna service, as it was started externally")
```

ray_on_golem/server/views.py
```diff
@@ -159,3 +166,26 @@ async def get_or_create_ssh_key(request):
...
+
+    if not request.app["self_shutdown"]:
+        shutdown_state = ShutdownState.NOT_ENABLED
+    elif await ray_service.get_non_terminated_nodes_ids():
+        shutdown_state = ShutdownState.CLUSTER_NOT_EMPTY
+    else:
+        shutdown_state = ShutdownState.WILL_SHUTDOWN
+
+    if shutdown_state == ShutdownState.WILL_SHUTDOWN:
+        logger.info("Received self shutdown request, exiting in 10 seconds...")
```
shadeofblue, line 185: again, a twice-hardcoded magic number...

shadeofblue (commented):
added my comments

ray_on_golem/client/client.py
approxit: Good point, added configurable and time based waiting.

ray_on_golem/client/client.py
approxit: There is a missing word, but I'm keeping symmetric logs text with line `Requesting webserver shutdown...` to keep logs more readable.

ray_on_golem/server/run.py
approxit: Well, good point, added.

But regarding `host` - I've never changed this from `0.0.0.0` to anything else in my whole career. Even more  - every time where host was not `0.0.0.0` it causes some problems. Having host in that way is enough for my taste.

shadeofblue (approved):
I believe it looks good now

PR #71, opened 2023-09-20
Dev cluster config

approxit:
What I've done:
- Created `golem-cluster-dev.yaml` file, with disabled registry stats and enabled project files sync
- Cleaned up some fields in cluster yaml files
- Added support for disabling registry stats on image resolution
- Added default value for `auth/ssh_user` field
- Fixed problem with `disable_existing_loggers` settings in logging facility, that made bunch of premature exit problems invisible
- Fixed problem with premature exit of managed yagna process due to `SIGINT` forwarding from parent to child subprocess, that made problems with clean `GolemNode` shutdown
- Fixed problem with premature webserver exit on self-shutdown, that made final exit logs not reachable
- Added more verbose logs of service lifetime
- Reformatted code

Notable remarks:
- Forcibly exiting webserver can leave `yagna` process running in the background
- Terminating nodes with `ray down` can have race condition with autoscaler in the background, and webserver self-shutdown can be called multiple times, exiting too early in result, witch raises ugly exception on `ray down`

### DIFF
--- a/ray_on_golem/server/services/yagna.py
+++ b/ray_on_golem/server/services/yagna.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import logging
+import os
 from asyncio.subprocess import Process
 from pathlib import Path
 from typing import Optional
@@ -65,6 +66,7 @@ class YagnaService:
             "run",
             stdout=asyncio.subprocess.DEVNULL,
             stderr=asyncio.subprocess.DEVNULL,
+            preexec_fn=os.setpgrp,  # https://stackoverflow.com/a/5446982/1993670
         )
 
         is_running = await self._wait_for_yagna_api()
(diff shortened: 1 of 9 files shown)


ray_on_golem/server/services/yagna.py
```diff
@@ -65,6 +66,7 @@ async def _run_yagna_service(self) -> None:
             "run",
             stdout=asyncio.subprocess.DEVNULL,
             stderr=asyncio.subprocess.DEVNULL,
```
lucekdudek, line 68: Can we still access yagna logs soemhow?

lucekdudek (approved):

ray_on_golem/server/services/yagna.py
approxit: Yes, `tail -n 50 -f ~/.local/share/yagna/yagna_rCURRENT.log`

shadeofblue (approved):

PR #80, opened 2023-09-22
Better subprocess management

approxit:
What I've done:
- Fixed problem with dangling reverse tunnel, as tunnel's parent subshell was terminated properly and not actual tunnel process
- Fixed quotation approach with `ProxyCommand`, as it causes problems with execution between subshell and direct exec approach
- Fixed hardcoded `root` username in few places around ssh usage
- Fixed typo from `tunel` to `tunnel`.
- Added more verbose logs to services
- Added warning logs when tunnel and yagna processes exit prematurely
- Reformatted the code

### DIFF
--- a/ray_on_golem/server/services/golem/golem.py
+++ b/ray_on_golem/server/services/golem/golem.py
@@ -66,10 +66,6 @@ class GolemService:
             "192.168.0.1/24"
         )  # will be retrieved from provider_config
         await self._golem.add_to_network(self._network)
-        self._allocation = await self._golem.create_allocation(
-            amount=1, network="goerli", autoclose=True
-        )
-        self._payment_manager = DefaultPaymentManager(self._golem, self._allocation)
 
         logger.info("Starting GolemService done")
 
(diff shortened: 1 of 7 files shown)


ray_on_golem/server/services/golem/golem.py
```diff
@@ -82,11 +78,14 @@ async def shutdown(self) -> None:
         """
         logger.info("Stopping GolemService...")
 
-        await self._payment_manager.terminate_agreements()
+        if self._payment_manager is None:
```
lucekdudek, line 81: It is not clear why `self._payment_manager is None` indicates `cluster was not started`

ray_on_golem/server/services/ray.py
```diff
@@ -187,34 +191,59 @@ async def _get_head_node(self) -> Node:
...
 
-    async def _stop_head_node_to_webserver_tunel(self) -> None:
-        process = self._head_node_to_webserver_tunel_process
+        logger.warning(f"Head node to webserver tunnel exited prematurely!")
 
-        if process is None:
+    async def _stop_head_node_to_webserver_tunnel(self) -> None:
+        process = self._head_node_to_webserver_tunnel_process
+
+        if process is None or process.returncode is not None:
```
lucekdudek, line 232: Why not ask for forgiveness instead of asking for permission here?
```python
try:
    self._head_node_to_webserver_tunnel_early_exit_task.cancel()
 ...
 ```

ray_on_golem/server/services/golem/golem.py
approxit: Well, it comes from the flow. `create_cluster` is the only place where `PaymentManager` is created, but at shutdown, we need take into the account, that server can be stopped before calling first `create_cluster`.

ray_on_golem/server/services/ray.py
approxit: The only reason is to have explicit logs. Same (or even more) applies to handling of the yagna process.

lucekdudek (approved):

PR #88, opened 2023-10-05
Managers support

approxit:
What I've done:
- Removed pipe-based `GolemService` in favor of manager-based `GolemService`
- Implemented baseline for future support of `node_config` sourced from worker yaml definition instead of provider yaml definition.
- Added upload of local golem-core files in `golem-cluster-dev.yaml`
- Added few market strategy parameters to to yaml for customizing market strategy

Notable remarks:
- `golem-cluster-dev.yaml` will expect a `golem-core-python` project directory next to `ray-on-golem` project directory in your env setup


golem-cluster-dev.yaml
```diff
@@ -30,7 +30,7 @@ provider:
     node_config:
       # if not provided, image_tag will be autodetected based on currently used python and ray versions
       # check available versions at https://registry.golem.network/explore/golem/ray-on-golem
-      image_tag: "lucjandudek/ray-on-golem:py3.10.12-ray2.3.1-coremain"
+      image_tag: "lucjandudek/ray-on-golem:managers"
```
lucekdudek, line 33: Better tag name would be welcome

examples/mandelbrot-golem-cluster.yaml
```diff
@@ -94,9 +94,9 @@ worker_start_ray_commands: [
 ]
 
 # Authentication credentials that Ray will use to launch nodes
-auth:
-  # TODO: try to remove explicit ssh_user field
-  ssh_user: 'root'
+# auth:
+  # Custom username for ssh
+  # ssh_user: "root"
```
lucekdudek, line 99: Do we want to keep this here?

approxit: This change was done only for the sake of keeping full example up to date with top level `golem-cluster.yaml`.

golem-cluster-dev.yaml
approxit: As dev cluster copies ray-on-golem and golem-core-python over the activities, there is no need for special image.
I'm defaulting this value to non-dev value, which will be latest release.

golem-cluster-dev.yaml
```diff
@@ -30,7 +30,7 @@ provider:
     node_config:
       # if not provided, image_tag will be autodetected based on currently used python and ray versions
       # check available versions at https://registry.golem.network/explore/golem/ray-on-golem
-      image_tag: "lucjandudek/ray-on-golem:py3.10.12-ray2.3.1-coremain"
+      image_tag: "golem/ray-on-golem:0.1.2-py3.10.13-ray2.3.1"
```
lucekdudek, line 33: Are you sure `golem/ray-on-golem:0.1.2-py3.10.13-ray2.3.1` has the same golem-core dependencies as the golem-core:main?

lucekdudek (approved):

PR #89, opened 2023-10-11
more sensible way to capture webserver output/error streams

shadeofblue:


lucekdudek (approved):

approxit (approved):
We're gonna work some more on logging, so this fix will not last long, but its fair enough right now.

PR #91, opened 2023-10-18
Better subprocesses logs

approxit:
What I've done:
- Reworked logging for subprocesses (webserver/yagna), following of the @shadeofblue idea:
  - stdout/stderr is redirected to debug file
  - If subprocess fail to start, last 50 lines of the log are printed to the user with mentioned file path to see full log.
- Made webserver to log with debug to console, and log with info to file
- Removed unneeded `log_level` parameter, as logs gathered intentionally unconditionally
- Fixed bug in `GolemService._apply_cost_management_hard_limits` that produced wrong log messages by applying plugins too early
- Added explicit guard if ray tried to spawn nodes before cluster bootstrap (happens if there are some leftovers from last webserver instance and its forceful exit)
- Reworded few logs
- Fixed problem with metadata missmatch in pip while installing Pillow package in `dds-with-ray.py` example.

Notable remarks:
- Redirecting stdout of the process to file is most reliable way to see debugs logs, as all non logging facility quirks are printed there (same for python and even yagna!)
- Having webserver to log with debug to console, and log with info to file is a great settlement between explicitly running webserver in development and implicitly running webserver by user

### DIFF
--- a/examples/dds-with-ray.py
+++ b/examples/dds-with-ray.py
@@ -4,7 +4,12 @@ import ray
 
 DEAL_COUNT = 50
 # endplay library dependency
-runtime_env = {"pip": ["endplay==0.4.11b0"]}
+runtime_env = {
+    "pip": [
+        "Pillow",  # explicit uppercased "Pillow", as pip complains about mismatched metadata in endplay dependency tree
+        "endplay",
+    ]
+}
 
 # Use default ray cluster or start a local one
 # Make sure endplay lib is installed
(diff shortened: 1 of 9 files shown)


examples/dds-with-ray.py
```diff
@@ -4,7 +4,12 @@
 
 DEAL_COUNT = 50
 # endplay library dependency
-runtime_env = {"pip": ["endplay==0.4.11b0"]}
```
mateuszsrebrny, line 7: why did we drop "==0.4.11b0"
I don't know what is in the newer versions, and I thought having the explicit version protects our example from potential future incompatibilities :thinking: in the future ?

approxit: Whoops! Forget in battle. We should pin that version.

PR #93, opened 2023-10-24
Binary dependencies and dynamic ray head ip

approxit:
What I've done:
- Added `yagna` and `websocat` python dependencies. As `yagna` is more version dependent, exact value is pinned
- Changed cluster yaml to use `$RAY_HEAD_IP` env variable (prepared by autoscaler alone)
- Minor log enchancements

Notable remarks:
- Last time we've tried using `$RAY_HEAD_IP` env variable it was not working, probably because of old approach with head on local machine. Now, it just works out of the box, duh.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,8 @@ aiohttp = "^3"
 requests = "^2"
 click = "^8"
 pydantic = "<2"
+websocat = "^1.12.0"
+golem-node = "0.12.2"
 
 setuptools = "*"
 
--- a/ray_on_golem/provider/node_provider.py
+++ b/ray_on_golem/provider/node_provider.py
@@ -212,20 +212,25 @@ class GolemNodeProvider(NodeProvider):
                         cli_logger.print("Starting webserver done")
                         return
                 else:
-                    cli_logger.abort(
-                        "Starting webserver failed!\nShowing last 50 lines from `{}`:\n{}",
-                        LOGGING_DEBUG_PATH,
-                        "".join(LOGGING_DEBUG_PATH.open("r").readlines()[-50:]),
-                    )
+                    with LOGGING_DEBUG_PATH.open("r") as file:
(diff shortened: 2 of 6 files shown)


ray_on_golem/provider/node_provider.py
```diff
@@ -212,20 +212,25 @@ def _start_webserver(
...
 
-            cli_logger.abort(
-                "Starting webserver failed! Deadline of `{}` reached.", RAY_ON_GOLEM_START_DEADLINE
-            )
+            with LOGGING_DEBUG_PATH.open("r") as file:
+                cli_logger.abort(
+                    "Starting webserver failed! Deadline of `{}` reached.\nShowing last 50 lines from `{}`:\n{}",
+                    RAY_ON_GOLEM_START_DEADLINE,
+                    LOGGING_DEBUG_PATH,
+                    "".join(file.readlines()[-50:]),
```
shadeofblue, line 232: the only issue I have here is that if the log is large, we'll need to read the whole just to show the last 50 lines ... maybe we could use an approach suggested here: https://stackoverflow.com/questions/46258499/how-to-read-the-last-line-of-a-file-in-python ?

shadeofblue (approved):
seems more or less okay, with one reservation (added as a comment)

pyproject.toml
```diff
@@ -16,6 +16,8 @@ aiohttp = "^3"
 requests = "^2"
 click = "^8"
 pydantic = "<2"
+websocat = "^1.12.0"
+golem-node = "0.12.2"
```
mateuszsrebrny, line 20: Reqc told me to use 0.13.1 as 12.3 has something for us (but it is not piped...)

approxit: Yup, we are aware of that request, yet pushing new yagna version to the users, without significant testing from our side is not a good idea from my side.

PR #99, opened 2023-10-25
fix the dependencies of installed packages, install only the required…

shadeofblue:
… components inside the image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "ray-on-golem"
-version = "0.1.3"
+version = "0.2.0-dev"
 description = "Golem Network integration with Ray"
 authors = ["Golem Factory <contact@golem.network>"]
 readme = "README.md"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -17,10 +16,21 @@ requests = "^2"
...
-golem-node = "0.12.2"
+golem-node = "0.12.3"
 
 setuptools = "*"
 
+[tool.poetry.group.ray.dependencies]
+python = "^3.8"
+ray = {version="^2.7.1", extras=["default"]}
+# golem-api-python = { path="../golem-core-python", develop=true }
+golem-api-python = { git="https://github.com/golemfactory/golem-core-python.git", branch="main" }
```
lucekdudek, line 27: What do we need `golem-api-python` on ray for?

shadeofblue: that's one of the dependencies of `ray-on-golem`...

approxit: Unfortunately, in current state imports requires golem core event for running inside of the provider.

lucekdudek (approved):

PR #110, opened 2023-11-08
Smoke tests

approxit:
What I've done:
- Added GH workflow that runs `ray up`, some example script and `ray down`
- Results are checked if worker nodes were symmetrically created and terminated
- Multiple attempts in `yagna payment fund`, as sometimes single run runs on timeouts
- Extended delay for self shutdown, as sometimes `10` was not enough.
- Usage of yagna in `0.13.1` version


.github/workflows/smoke_tests.yml
```diff
@@ -23,20 +23,20 @@ jobs:
       - name: Install dependencies
         run: poetry install --no-ansi
 
-      - name: Prepare `golem-cluster-local.yaml`
+      - name: Prepare `golem-cluster.local.yaml`
```
shadeofblue, line 26: `golem-cluster.local.yaml` is supposed to be a file that actually exists locally before `utils.apply_overrides` is called - and which contains local changes for a specific user - by default, that file gets included in the `apply_overrides` operation as the bottom-most override (the last one that gets applied) ... so, probaby the name of the _output_ file should be different, e.g. `golem-cluster.tests.yaml` ... or something ... (mostly, something other that `.local.yaml`)...

shadeofblue:
I believe this pull request should be targeted towards `develop` first ... because otherwise, the changes from develop to main will get squashed... which we don't want...

pyproject.toml
```diff
@@ -17,7 +17,7 @@ requests = "^2"
 click = "^8"
 pydantic = "<2"
 websocat = "^1.12.0"
-golem-node = "0.12.3"
+golem-node = "0.13.1"
```
shadeofblue, line 20: whoah, do we want to upgrade yagna to `0.13.1` already?

approxit: Reverted

shadeofblue (approved):

PR #116, opened 2023-11-09
retry SSH connections on failure

shadeofblue:


ray_on_golem/server/main.py
```diff
@@ -52,7 +52,7 @@ def main(port: int, self_shutdown: bool, registry_stats: bool):
         web.run_app(
             app,
             port=app["port"],
-            print=None,
+            print=logger.info,
```
lucekdudek, line 55: How much does it log? Maybe `debug` would be better here?

ray_on_golem/server/services/golem/golem.py
```diff
@@ -196,30 +196,48 @@ async def _verify_ssh_connection(
...
 
-        logger.info(msg)
+        while retry > 0:
+            try:
+                await check()
+                break
+            except Exception as e:
+                retry -= 1
+                if retry:
+                    logger.warning(f"{str(e)}, retrying {retry}...")
```
lucekdudek, line 233: This can be `logger.debug` we were already considering making all failed ssh of `DEBUG` level

ray_on_golem/server/main.py
shadeofblue: hmm.... if it's something that gets _printed_ normally, I don't think there's much... and I do believe what _is_ being output would constitute regular info

ray_on_golem/server/services/golem/golem.py
shadeofblue: hmm... why? it does sound "warning"-like ... I mean, it _should_ work normally... and maybe it _is_ a sign of something amiss?

lucekdudek: Because there are many of them already and with retries there will be more.

ray_on_golem/server/main.py
lucekdudek: ok

ray_on_golem/server/services/golem/golem.py
shadeofblue: but it's still abnormal... imvho...

lucekdudek (approved):

PR #128, opened 2023-11-27
Intermediate node states

approxit:
What I've done:
* Introduced intermediate node state support in ray's cluster
* New node is added to cluster instantly, instead of at the very end of activity creation
* Progress in activity creation will be reported to cluster state
* Node provider while requesting new nodes will report state changes
* IP / ssh proxy command / activity will be `None` until activity will be available for the node

Notable remarks:
* `get_cluster_state` endpoint will enable simplification in `RayOnGolemClient`


shadeofblue (approved):
looks good as far as I could tell, though there's quite a lot happening there :)

lucekdudek (approved):

PR #139, opened 2023-12-05
Add pillow to default image

lucekdudek:
- [x] Added curl and pillow to `Dockerfile`
 - [x] Update image tag in mandelbrot cluster yaml
 - [x] Added `--output-file` option to mandelbrot example


shadeofblue:
@lucekdudek whereas I have nothing against such change, question still remeains, why are we also adding the output path there?

shadeofblue:
@mateuszsrebrny @lucekdudek plus, didn't we want to preserve the yaml to still show how you'd ensure that a library is added?

lucekdudek:
> @lucekdudek whereas I have nothing against such change, question still remeains, why are we also adding the output path there?

I decided to revert this change after all

lucekdudek:
> @mateuszsrebrny @lucekdudek plus, didn't we want to preserve the yaml to still show how you'd ensure that a library is added?



> @mateuszsrebrny @lucekdudek plus, didn't we want to preserve the yaml to still show how you'd ensure that a library is added?

Only unique feature are `initialization_commands`
```yaml
initialization_commands: [
    "pip install pillow",
]
```
and we already describe them in the main yaml file.

@mateuszsrebrny what do you think?

mateuszsrebrny:
> > @mateuszsrebrny @lucekdudek plus, didn't we want to preserve the yaml to still show how you'd ensure that a library is added?
> 
> Only unique feature are `initialization_commands`
> 
> ```yaml
> initialization_commands: [
>     "pip install pillow",
> ]
> ```
> 
> and we already describe them in the main yaml file.
> 
> @mateuszsrebrny what do you think?

hm... not sure - my first thought is that the yaml allowing an example to run is also part of this example
so even if it differs only slightly from the default yaml as a user I would prefer to simply download it, run it and be able to see that the example works... I wouldn't want to read comments about the needed packages in the code and try to modify the yaml to support them...

sounds sensible?
if you disagree lets talk in the morning, asynchronously I won't come up with anything better ;)

shadeofblue:
> sounds sensible? if you disagree lets talk in the morning, asynchronously I won't come up with anything better ;)

yeah... and again, I'd treat removal of the yaml file as a separate step (feature, so to say) from adding `pillow` to the image...

approxit (approved):

lucekdudek:
@mateuszsrebrny @shadeofblue 
Done, please take a look

PR #141, opened 2023-12-05
Onboarding message on mainnet

approxit:
What I've done:
- If `polygon` or `mainnet` is used as `payment_network` in yaml config, onboarding message will be printed at the beginning of the `ray up`.
- `yagna payment fund` is called at create_cluster state, regardless if webserver is managing yagna or not.

Notable remarks:
- We have still problem with allocation exceptions that are not propagating through the stack of managers / background tasks. This is kinda complex, so we need to design neat solution for it.

### DIFF
--- a/ray_on_golem/provider/node_provider.py
+++ b/ray_on_golem/provider/node_provider.py
@@ -30,7 +30,14 @@ from ray_on_golem.utils import (
 )
 
 LOG_GROUP = "Ray On Golem"
-MAINNET_PAYMENT_NETWORKS = ("mainnet", "polygon",)
+
+PAYMENT_NETWORK_MAINNET = "mainnet"
+PAYMENT_NETWORK_POLYGON = "polygon"
+
+ONBOARDING_MESSAGE = {
+    PAYMENT_NETWORK_MAINNET: "Running Ray on Golem on the Ethereum Mainnet requires GLM and ETH tokens.",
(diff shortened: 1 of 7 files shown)


lucekdudek (approved):

ray_on_golem/provider/node_provider.py
```diff
@@ -309,3 +320,27 @@ def _stop_webserver(ray_on_golem_client: RayOnGolemClient) -> None:
...
+
+    def _print_mainnet_onboarding_message(self, yagna_payment_status_output: str) -> None:
+        if self._payment_network not in MAINNET_PAYMENT_NETWORKS:
+            return
+
+        cli_logger.newline()
+
+        with cli_logger.indented():
+            cli_logger.print(
+                "Running Ray on Golem on the mainnet requires GLM and MATIC tokens on the Polygon blockchain"
```
shadeofblue, line 332: well, we're checking for `mainnet` but then only seem to support Polygon ...

we should either:
* only print this message if the user is indeed using Polygon and print a different message for actual _mainnet_ users
* or ensure that our onboarding also supports the Ethereum Mainnet and differentiate the message depending on the actual blockchain used by the user

ray_on_golem/server/models.py
```diff
@@ -92,6 +92,12 @@ class CreateClusterRequestData(ProviderConfigData):
     pass
 
 
+class CreateClusterResponseData(BaseModel):
+    is_cluster_just_created: bool
```
shadeofblue, line 96: the name of this variable is very awkward imo... 

```suggestion
    initial_startup: bool
```

?

mateuszsrebrny (approved):
:+1:

PR #152, opened 2023-12-18
fix `yagna payment fund` deadline to allow for actual fund timeouts ;)

shadeofblue:


shadeofblue:
I believe it's not needed anymore :/

PR #156, opened 2023-12-19
`[5/9] Starting VM container` message (instead of `[5/9] Starting activity`)

mateuszsrebrny:
`Starting activity` sounds weird after `Creating activity` and `Adding activity to internal VPN` :)

### DIFF
--- a/ray_on_golem/server/services/golem/golem.py
+++ b/ray_on_golem/server/services/golem/golem.py
@@ -194,8 +194,8 @@ class GolemService:
         deploy_args = {"net": [self._network.deploy_args(ip)]}
         await context.deploy(deploy_args, timeout=timedelta(minutes=5))
 
-        await add_state_log("[5/9] Starting container...")
-        logger.info(f"Starting a VM container on {provider_desc}, {ip=}, {activity=}")
+        await add_state_log("[5/9] Starting VM container...")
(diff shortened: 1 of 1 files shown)


ray_on_golem/server/services/golem/golem.py
```diff
@@ -194,8 +194,8 @@ async def _start_activity(
         deploy_args = {"net": [self._network.deploy_args(ip)]}
         await context.deploy(deploy_args, timeout=timedelta(minutes=5))
 
-        await add_state_log("[5/9] Starting activity...")
-        logger.info(f"Starting activity on {provider_desc}, {ip=}, {activity=}")
+        await add_state_log("[5/9] Starting container...")
+        logger.info(f"Starting container on {provider_desc}, {ip=}, {activity=}")
```
shadeofblue, line 198:
```suggestion
        logger.info(f"Starting a VM container on {provider_desc}, {ip=}, {activity=}")
```

shadeofblue (approved):
one suggestion though ;)

ray_on_golem/server/services/golem/golem.py
mateuszsrebrny: added VM to both log lines, but didn't add `a` - it would look weird as in no other steps we have articles (and I don't think adding them everywhere is a good idea (it would mess up the output IMO)

PR #160, opened 2024-01-02
Change default network mask to 16

lucekdudek:


ray_on_golem/server/services/golem/golem.py
```diff
@@ -53,7 +53,7 @@ async def init(self, yagna_appkey: str) -> None:
         await self._golem.start()
 
         self._network = await self._golem.create_network(
-            "192.168.0.1/24"
+            "192.168.0.1/16"
```
shadeofblue, line 56: why not 8, as discussed?

shadeofblue (approved):
was there a problem with setting the network address to `192.168.0.0/8`?

anyway, 16bit should also be quite sufficient...

ray_on_golem/server/services/golem/golem.py
lucekdudek: `192.168.0.0` doesn't allow `8` and `16` is more then enough

PR #174, opened 2024-02-02
Canceling pending node creation

approxit:
What I've done:

* Updated the project to use newest golem-core version from git (until new release)
* Added proposal expiration from buffers, according to new `Proposal.get_expiration_date()` method
* Fixed warnings about legacy fields on `ray up`, by removing them at config building stage
* Fixed problem with hanging SSH subprocesses, by disabling asking for plain password

Notable remarks:
* As built-in providers have built-in support for loading their default values (that's why they don't need to define bunch of empty fields like `file_mounts`, etc), external ones needs to load their defaults by themselves. Its an opportunity to introduce proper minimal yaml requirements.

### DIFF
--- a/ray_on_golem/server/services/ray.py
+++ b/ray_on_golem/server/services/ray.py
@@ -6,7 +6,7 @@ from functools import partial
 from pathlib import Path
 from typing import Any, Dict, Iterator, List, Optional, Tuple
 
-from golem.utils.asyncio import create_task_with_logging
+from golem.utils.asyncio import create_task_with_logging, ensure_cancelled_many
 from golem.utils.logging import get_trace_id_name
 from ray.autoscaler.tags import NODE_KIND_HEAD, TAG_RAY_NODE_KIND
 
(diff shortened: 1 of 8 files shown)


ray_on_golem/server/services/ray.py
```diff
@@ -142,28 +147,33 @@ async def request_nodes(
...
-            node_config=self._provider_config.node_config,
-            public_ssh_key=self._ssh_public_key,
-            ssh_user=self._ssh_user,
-            ssh_private_key_path=self._ssh_private_key_path,
-            total_budget=self._provider_config.total_budget,
-            payment_network=self._provider_config.payment_network,
-            subnet_tag=self._provider_config.subnet_tag,
-            add_state_log=partial(self._add_node_state_log, node_id),
-        )
+        try:
```
shadeofblue, line 150: why are we wrapping all the code here in a try block instead of just wrapping the `create_activity` call?

shadeofblue (commented):
hmm, in the branch where I tried to abort `ray up` on an allocation failure, I remember having to update also the `get_cluster_data` view and RayService's `get_cluster_data` method to correctly handle the error ... is that no longer needed?

attaching the link to those changes just for reference:

https://github.com/golemfactory/ray-on-golem/compare/main...blue/allocation-abort#diff-9c71e4ca4fcb522d17af9315286cd4553acb52cf8bff461ee05079a0b5a1184eR201

ray_on_golem/server/services/ray.py
approxit: Well, in this case, `_create_node()` is scheduled as a task, and is run in the background. There is no point in the code to await that task, hence it needs to wrap itself with `try/catch` instead of relying on the caller.

PR #175, opened 2024-02-06
Canceling pending node creation

approxit:
On top of https://github.com/golemfactory/ray-on-golem/pull/174.

What I've done:
* Fixed removing warning about legacy fields
* Made GolemNodeProvider handle error propagation

Notable remarks:
* The is currently no option to disable legacy fields warning with the external provider without any addition to the original yaml.

### DIFF
--- a/ray_on_golem/server/services/golem/golem.py
+++ b/ray_on_golem/server/services/golem/golem.py
@@ -2,14 +2,13 @@ import asyncio
 import hashlib
 import logging
 from collections import defaultdict
-from datetime import timedelta
+from datetime import datetime, timedelta, timezone
 from functools import partial
 from pathlib import Path
 from typing import Awaitable, Callable, Dict, Optional, Tuple
 
 from golem.managers import (
     BlacklistProviderIdPlugin,
-    Buffer,
     DefaultAgreementManager,
     DefaultProposalManager,
     MapScore,
@@ -17,13 +16,14 @@ from golem.managers import (
     NegotiatingPlugin,
     PayAllPaymentManager,
     PaymentPlatformNegotiator,
+    ProposalBuffer,
+    ProposalScoringBuffer,
     RefreshingDemandManager,
-    ScoringBuffer,
     WorkContext,
 )
 from golem.node import GolemNode
 from golem.payload import PaymentInfo
-from golem.resources import Activity, Network, ProposalData
+from golem.resources import Activity, Network, Proposal, ProposalData
 from yarl import URL
 
 from ray_on_golem.server.models import NodeConfigData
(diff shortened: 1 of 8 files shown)


ray_on_golem/server/settings.py
```diff
@@ -59,7 +59,7 @@
             "propagate": False,
         },
         "ray_on_golem": {
-            "level": "DEBUG",
+            "level": "INFO",
```
shadeofblue, line 62: why?

shadeofblue (approved):
looks okay

please comment on the log level change, though :)

ray_on_golem/server/services/golem/golem.py
```diff
@@ -208,6 +214,11 @@ async def _create_stack(
 
         return stack
 
+    async def _get_proposal_expiration(self, proposal: Proposal) -> timedelta:
+        await proposal.get_data()
+
+        return proposal.get_expiration_date() - datetime.now(timezone.utc)
```
lucekdudek, line 221: I know it is golem-core but wouldn't it be better if `get_expiration_date` be asyncio function and handle `await proposal.get_data()`?

ray_on_golem/server/services/golem/golem.py
```diff
@@ -376,7 +387,12 @@ async def _create_activity(
         logger.info(f"Creating new activity...")
 
         await add_state_log("[1/9] Getting agreement...")
-        agreement = await stack.agreement_manager.get_agreement()
+
+        try:
+            agreement = await stack.agreement_manager.get_agreement()
+        except Exception as e:
+            logger.error(f"Creating new activity failed with `{e}`")
+            raise RuntimeError(e) from e
```
lucekdudek: Is `RuntimeError` the best here?

ray_on_golem/server/settings.py
approxit: Change reverted. :)

ray_on_golem/server/services/golem/golem.py
approxit: It depends. This function follows a pattern that exists in golem-core - helper functions expect to have `.get_data()` called elsewhere. I decided to keep it consistent instead of introducing something different.

ray_on_golem/server/services/golem/golem.py
approxit: This comment was created after the change. :D

It was changed to `GolemException` for now.

ray_on_golem/server/services/golem/golem.py
approxit: We decided that usage of `async` is desired as in `Proposal` we have few methods in that way.

lucekdudek (approved):

Issue #173, opened 2024-01-30
ray-on-golem start / stop


shadeofblue:
![image](https://github.com/golemfactory/ray-on-golem/assets/5002408/edf35534-2e2d-41ee-928f-cd4ee33bec98)

PR #177, opened 2024-02-09
Proper stop handling

shadeofblue:
https://github.com/golemfactory/ray-on-golem/issues/173

### DIFF
--- a/ray_on_golem/ctl/ctl.py
+++ b/ray_on_golem/ctl/ctl.py
@@ -110,7 +110,9 @@ class RayOnGolemCtl:
             return
 
         self._output_logger.info(
-            f"{'Requesting' if not force_shutdown else 'Forcing' } webserver shutdown..."
+            "{verb} webserver shutdown...".format(
+                verb='Requesting' if not force_shutdown else 'Forcing'
+            )
         )
 
         shutdown_state = self._client.shutdown_webserver(
--- a/ray_on_golem/ctl/ctl.py
+++ b/ray_on_golem/ctl/ctl.py
@@ -37,7 +37,7 @@ class RayOnGolemCtl:
             if webserver_status.shutting_down:
                 self.wait_for_shutdown()
             else:
-                self._output_logger.info("Not starting webserver, as it's already running")
+                self._output_logger.info("Not starting the webserver, as it's already running")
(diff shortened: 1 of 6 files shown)


lucekdudek (approved):

ray_on_golem/ctl/ctl.py
```diff
@@ -98,30 +97,36 @@ def start_webserver(
...
-                self._output_logger.info(
-                    "Not stopping the webserver, as it's already shutting down"
-                )
+                self._output_logger.info("The webserver is already shutting down.")
 
             return
 
-        self._output_logger.info("Requesting webserver shutdown...")
+        self._output_logger.info(
+            f"{'Requesting' if not force_shutdown else 'Forcing' } webserver shutdown..."
```
approxit, line 113: I prefer `.format()` instead of f-string for that kind of complex formatting.

ray_on_golem/server/settings.py
```diff
@@ -31,7 +31,7 @@
 URL_TERMINATE_NODE = "/terminate"
 URL_GET_SSH_PROXY_COMMAND = "/ssh_proxy_command"
 URL_GET_OR_CREATE_DEFAULT_SSH_KEY = "/ger_or_create_default_ssh_key"
-URL_SELF_SHUTDOWN = "/self_shutdown"
+URL_SHUTDOWN = "/shutdown"
```
approxit, line 34: If we are changing terms and removing the `self` part, let's do it globally.

shadeofblue: if you mean updating every piece of code to remove the word "self" from there, I think it's too much here -> this can be done with a separate pull request if indeed needed - what I wanted to have here is a cleaner API (where the word "self" doesn' really give any meaning) but I don't see how that implies fixing _all_ the code that contains it right now ...

shadeofblue: I mean, of course it can be done, I just think that putting such remarks on pull requests doesn't serve us well and hinders our progress more than helps it...

shadeofblue: anyway, @approxit I just updated the name of the view to match the URL -> other places actually mean what they should -> iow, they enable ability to _self_ shutdown

approxit (approved):

PR #178, opened 2024-02-13
Stop subprocesses on webserver stop

lucekdudek:
[x] Change how run_subprocess util function works. Now it accepts additional argument which is only needed for yagna subprocess
[x] Added RandomScore to manager stack to slightly deviate order of processing proposals

### DIFF
--- a/ray_on_golem/utils.py
+++ b/ray_on_golem/utils.py
@@ -15,16 +15,13 @@ async def run_subprocess(
     *args,
     stderr=asyncio.subprocess.DEVNULL,
     stdout=asyncio.subprocess.DEVNULL,
-    disable_signal_propagation=False,
+    detach=False,
 ) -> Process:
-    # If the process lifetime will be fully managed, we need to disable signal propagation from
-    # parent to child process https://stackoverflow.com/a/5446982/1993670
-    kwargs = {"preexec_fn": os.setpgrp} if disable_signal_propagation else {}
     process = await asyncio.create_subprocess_exec(
         *args,
         stderr=stderr,
         stdout=stdout,
-        **kwargs,
+        start_new_session=detach,
     )
 
     return process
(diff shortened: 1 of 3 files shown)


ray_on_golem/utils.py
```diff
@@ -12,22 +12,26 @@
...
+    # If the process lifetime will be fully managed, we need to disable signal propagation from
+    # parent to child process https://stackoverflow.com/a/5446982/1993670
+    kwargs = {"preexec_fn": os.setpgrp} if disable_signal_propagation else {}
     process = await asyncio.create_subprocess_exec(
         *args,
         stderr=stderr,
         stdout=stdout,
-        # As this process lifetime will be fully managed, we need to disable signal propagation
-        # from parent to child process https://stackoverflow.com/a/5446982/1993670
-        preexec_fn=os.setpgrp,
```
approxit, line 23: This approach was vailable for older versions of Python. For our version let's use the `start_new_session` kwargument instead. This can be a name for the argument instead of `disable_signal_propagation`.

ray_on_golem/utils.py
```diff
@@ -12,22 +12,26 @@
...
+        **kwargs,
```
shadeofblue, line 27: re @approxit ^^^ or, rather, name the parameter `detach` as it's self explanatory and pass it explicitly as:

```suggestion
        start_new_session=detach,
```

pyproject.toml
```diff
@@ -13,6 +13,7 @@ ray = {version="^2.7.1", extras=["default"]}
 golem-core = { git="https://github.com/golemfactory/golem-core-python.git", branch="main" }
 #golem-core = "^0.4.0"
 aiohttp = "^3"
+appdirs = "^1.4.4"
```
shadeofblue, line 16: https://github.com/golemfactory/ray-on-golem/pull/179

approxit (approved):

shadeofblue (approved):

PR #181, opened 2024-02-15
Support `node_config` per node type in yaml

approxit:
What I've done:
- Moved `subnet_tag` to be part of the `node_config` instead of `provider.parameters` as the subnet now can differ from node type to node type.
- Made manager stacks to be based on `available_node_types.<node type>.node_config` and not on `provider.parameters.node_config`.
- Made config bootstrapping to populate `available_node_types.<node type>.node_config` with defaults from `provider.parameters.node_config`, so common settings declared in `provider.parameters.node_config` will be shared/copied over to all node types.
- Removed explicit server shutdown in `NodeProvider.terminate_node` in favor of closing unused manager stack on the webserver side while calling that function.
- Pulled out the payment manager from the manager stack, as it is common for the whole cluster, not per manager stack.
- Made `GolemService` to track which activity uses which stack, and expose activity_id instead of raw activity to help with tracking safety.
- Made `network_stats` entry point to be based on a specific node type, as each node type could have different market settings
- Updated ray version to `2.9.2` and cleaned up / normalized the dependencies.
- Updated gvmi image

Notable remarks:
- Somehow, once again, exiting the webserver waits to timeout instead of exiting right away... this should be somehow investigated and fixed.

### DIFF
--- a/golem-cluster.yaml
+++ b/golem-cluster.yaml
@@ -25,6 +25,7 @@ provider:
     # Maximum amount of GLMs that's going to be spent for the whole cluster
     total_budget: 5
 
+    # Common parameters for all node types. Can be overriden in available_node_types
     node_config:
       # Parameters for golem demands (same for head and workers)
       demand:
@@ -100,7 +101,9 @@ available_node_types:
     # The node type's CPU and GPU resources
     resources: {}
 
-    node_config: {} # TODO: Demand description here
+    # Additional parameters specific for this node type added on top of node_config from provider.parameters.node_config
+    node_config: {}
+
   ray.worker.default:
     min_workers: 1
     max_workers: 10
--- a/golem-cluster.yaml
+++ b/golem-cluster.yaml
@@ -103,6 +103,9 @@ available_node_types:
 
     # Additional parameters specific for this node type added on top of node_config from provider.parameters.node_config
     node_config: {}
+    #node_config:
+    #  demand:
+    #    min_mem_gib: 10
 
   ray.worker.default:
     min_workers: 1
@@ -111,10 +114,10 @@ available_node_types:
     node_config: {}
 
 # List of commands that will be run to initialize the nodes (before `setup_commands`)
+initialization_commands: []
 #initialization_commands: [
 #  "pip install endplay",
 #]
-initialization_commands: []
 
 # List of shell commands to run to set up nodes
 setup_commands: []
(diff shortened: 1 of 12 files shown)


approxit:
After a talk, I missed that stacks should be stopped at the end of the cluster, not after the last node in the stack. The last commit will reflect that change.

lucekdudek:
> * Pulled out the payment manager from the manager stack, as it is common for the whole cluster, not per manager stack.

I don't like how this change is implemented as it moves as further away from ability to support multiple clusters with a single webserver.

golem-cluster.yaml
```diff
@@ -100,7 +101,9 @@ available_node_types:
     # The node type's CPU and GPU resources
     resources: {}
 
-    node_config: {} # TODO: Demand description here
+    # Additional parameters specific for this node type added on top of node_config from provider.parameters.node_config
+    node_config: {}
```
lucekdudek, line 105: Can we add an example `yaml` that is showcasing usage of different configuration for both stacks?

approxit:
> > * Pulled out the payment manager from the manager stack, as it is common for the whole cluster, not per manager stack.
> 
> I don't like how this change is implemented as it moves as further away from ability to support multiple clusters with a single webserver.

While making this PR I realized that we need to significantly redesign Ray/Golem services to truly support multiple clusters with a good quality of the interfaces. Besides, the payment manager can't be a part of a stack, as we need a single instance for multiple stacks.

shadeofblue:
please rebase onto / merge the latest develop

ray_on_golem/network_stats/main.py
```diff
@@ -14,13 +14,33 @@
...
+        )
+
+    return value
+
+
 @click.command(
     name="network-stats",
     short_help="Run Golem Network statistics.",
-    help="Run Golem Network statistics based on given cluster config file.",
+    help="Run Golem Network statistics based on given cluster config file and node type.",
```
shadeofblue, line 39: hmm... shouldn't network stats analyze _all_ the node types in the yaml?

ray_on_golem/network_stats/main.py
```diff
@@ -14,13 +14,33 @@
 from ray_on_golem.server.settings import DEFAULT_DATADIR, YAGNA_PATH, get_logging_config
 
 
+def validate_config_file(ctx, param, value):
```
shadeofblue, line 17: don't we wish to have these on the model validation level?

shadeofblue (approved):
anyway, looks good as far as the changes themselves are concerned

PR #182, opened 2024-02-16
update the dependencies, bump ray to 2.9

shadeofblue:


approxit (changes requested):
We need to update the used gvmi image too!

shadeofblue:
> We need to update the used gvmi image too!

updated

PR #188, opened 2024-02-21
Suggested heads support

approxit:
What I've done:
- Introduced a new demand manager for combining other demand managers (**I'm open to better names**)
- Made that in case of head node request, suggested nodes will be added to demands and scored better
- Network stats will use suggested nodes too, as it can impact the number of providers
- Removed explicit fields from ManagerStack in favor of just a collection to start/stop children in order, proving that Manager Stack is a blind alley
- Small yaml cleanup, adding explicit head node pointer for the sake of network stats logic

Notable remarks:
- This PR should be checked with golem-core changes: https://github.com/golemfactory/golem-core-python/compare/main...approxit/suggested-heads
- Suggested heads right now are only available on the mainnet.
- As the ManagerStack is still considered as a draft, it proved that it can only solve the start/stop problem, and can't solve the problem of how to access its children / partial definitions. `stack._managers[-1]` is a way with no exit. Another SDK API concept with stages should solve this issue.

### DIFF
--- a/golem-cluster.mini.yaml
+++ b/golem-cluster.mini.yaml
@@ -15,10 +15,6 @@ provider:
       demand:
         image_tag: "golem/ray-on-golem:0.7.0-py3.10.13-ray2.7.1"
 
-file_mounts: {
-  # <remote_path>: <local_path>
-}
-
 # Tells the autoscaler the allowed node types and the resources they provide
 available_node_types:
   ray.head.default:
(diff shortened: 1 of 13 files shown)


shadeofblue:
please create a complementary pull request for `golem-core-python`, thank you :)

approxit:
After additional consideration, we decided that suggested heads will be implemented a little more genetic, with `priority_head_subnet_tag` parameter in yaml. This will be used only in case of head node deployment, so this param can be shared in all `available_node_types`. Additional demand and scoring will be used in setup, only for the head node and when `priority_head_subnet_tag` is truthy.

By default `ray-on-golem-heads` subnet is used both for the mainnet and testnet. As providers are available only on the mainnet, it is acceptable for us to have the same configuration for the testnet anyways.
For dev purposes `sdk` subnet is used for `priority_head_subnet_tag`.

In addition, I've fixed how stacks are cached, as we need to distinguish now the head node demand from worker demand. Both could have the same configs, but context matters from now on.

shadeofblue (approved):
lgtm

Issue #197, opened 2024-02-26
Terminate node does not cancel its pending creation


approxit:
By working with AR, we encountered a problem: after stack removal, the flow of creating a new node still runs. That run fails due to a lack of existing state of managers etc.

`RayService` needs to track creation tasks per node id and cancel them when node termination is requested.

mateuszsrebrny:
done :)

PR #198, opened 2024-02-27
Goth examples on CI

approxit:
What I've done:
- Added all non-outbound vm examples on CI to be run nightly at 2 AM
- Removed default usage of outbound pypi URL (to ease example running)
- Fixed https://github.com/golemfactory/ray-on-golem/issues/197
- Made SSH check fails able to retry activity creation
- Removed hostname resolution dependency in `simple-task` and `simple-actor` examples, as sometimes goth instances had problems with their stability
- Fixed outdated `node_config` layout in mendelbrot yaml example
- Made `node_config` raise errors on unrecognized fields (to mitigate the previous element from this list)
- Fixed CI warnings about outdated node versions, by bumping versions of a few GHA steps
- Made yagna running checks explicit about the used URL
- Fixed problems with failing yagna funds on CI, by making its timeout a little bigger

Notable remarks:
- We decided that examples will be run not on pull requests and as nighty due to current problems with ssh stability


golem-cluster.yaml
```diff
@@ -35,7 +35,7 @@ provider:
         # List of urls which will be added to the Computation Manifest
         # Requires protocol to be defined in all URLs
         # If not provided demand will not use Computation Manifest
-        outbound_urls: ["https://pypi.dev.golem.network"]
+        # outbound_urls: ["https://pypi.dev.golem.network"]
```
lucekdudek, line 38: Why this change? Don't we use overwrite anyway?

golem-cluster.override.2-image.yaml
```diff
@@ -2,4 +2,4 @@ provider:
   parameters:
     node_config:
       demand:
-        image_tag: "blueshade/ray-on-golem:0.8.0-py3.10.13-ray2.9.2"
+        image_tag: "golem/ray-on-golem:0.8.0-py3.10.13-ray2.9.2"
```
lucekdudek, line 5: I think this is important for pre release versions to be not from the `golem` user registery.

pyproject.toml
```diff
@@ -9,9 +9,9 @@ packages = [{include = "ray_on_golem"}]
 [tool.poetry.dependencies]
 python = "^3.8.1"
 
-ray = {version="^2.9.2", extras=["default"]}
-#golem-core = {path="../golem-core-python", develop=true}
-#golem-core = {git="https://github.com/golemfactory/golem-core-python.git", branch="main"}
+ray = {version = "^2.9.2", extras = ["default"]}
+#golem-core = {path = "../golem-core-python", develop = true}
+#golem-core = {git = "https://github.com/golemfactory/golem-core-python.git", branch = "main"}
```
lucekdudek, line 14: My formatter would also change `#golem-core` into `# golem-core`. What are you using?

pyproject.toml
```diff
@@ -56,6 +57,11 @@ dpath = "^2.1.6"
 yamlpath = "^3.8.1"
 pytest = "^7.4.3"
 
+[[tool.poetry.source]]
+name = "test"
```
lucekdudek, line 61: Maybe `test-pypi` ?

.github/workflows/start-goth.sh
```diff
@@ -0,0 +1,55 @@
...
+rm -rf .envs/goth
+python -m venv .envs/goth
+source .envs/goth/bin/activate
+
+echo INSTALLING TOOLS
+python -m pip install --upgrade pip
+python -m pip install --upgrade setuptools wheel
+
+echo INSTALLING DEPENDENCIES
+python -m pip install goth==0.15.11 pytest pytest-asyncio pexpect
```
lucekdudek, line 24: Maybe we can move the goth version to params so it is easier to upgrade (or even run on multiple versions) later?

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,92 @@
...
+            args: 2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae
+
+          - example_name: mandelbrot_image
+            base_yaml_path: examples/mandelbrot-golem-cluster.yaml
+
+#          - example_name: outbound
+#            base_yaml_path: examples/outbound.yaml
+#
+#          - example_name: outbound_tester
+#            base_yaml_path: examples/outbound_tester.yaml
```
lucekdudek, line 29: Add a reason comment why those are disabled pls

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Collects logs
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: logs-example-${{ matrix.example_name }}
+          path: |
+            /root/.local/share/ray_on_golem/webserver_debug.log
+            /root/.local/share/ray_on_golem/yagna.log
+            /tmp/goth-tests
```
lucekdudek, line 92: Shouldn't we stop goth after all of this?

pyproject.toml
approxit: Well, PyCharm does not use space there. It would be great if we could use autoformat here...

.github/workflows/integration_tests.yml
approxit: Do you have any example for stopping goth?

lucekdudek:
```yaml
    - name: Remove Docker containers
      shell: bash
      if: always()
      continue-on-error: true
      run: docker rm --force $(docker ps -a -q)

    - name: Stop goth network
      shell: bash
      if: always()
      continue-on-error: true
      run: kill $(ps -aux | grep "[g]oth start" | awk '{print $2}')
```

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,96 @@
...
+          - example_name: simple-actor
+
+          - example_name: simple-task
+
+          # Commented out as Goth does not support outbound yet
+          # - example_name: dds-with-ray
+          #  - example_name: outbound
+          #    base_yaml_path: examples/outbound.yaml
+          #  - example_name: outbound_tester
+          #    base_yaml_path: examples/outbound_tester.yaml
```
lucekdudek, line 33: Formatting looks off
```python
# - example_
... vs
#  - example_
```

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,96 @@
...
+
+jobs:
+  examples:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - example_name: calculate_pi
+
+
```
lucekdudek, line 17: Extra empty line

.github/workflows/stop-goth.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+echo STOP GOTH
+kill $(ps -aux | grep "[g]oth start" | awk '{print $2}')
+
+echo STOP DOCKER CONTAINERS
+docker rm --force $(docker ps -a -q)
```
lucekdudek, line 6: I wonder if we can get PID from this
```
python -m goth start .envs/goth/assets/goth-config.yml &
GOTH_PID=$!
```

Or if there is a symetric command to the `goth start`

pyproject.toml
```diff
@@ -23,15 +23,16 @@ appdirs = "^1.4.4"
 
 websocat = "^1.12.0"
 golem-node = "0.14.0"
+#golem-node = {version = "0.15.0a18", source = "test"}
```
lucekdudek, line 26: source = "test-pipy"

lucekdudek (approved):

PR #199, opened 2024-02-28
add the actual hash cracker example running on ray

shadeofblue:


examples/hash_cracker_ray.py
```diff
@@ -0,0 +1,142 @@
...
+        range(ord("A"), ord("Z") + 1),
+        range(ord("0"), ord("9") + 1),
+        range(ord(" "), ord("/") + 1),
+        range(ord(":"), ord("@") + 1),
+        range(ord("["), ord("`") + 1),
+        range(ord("{"), ord("~") + 1),
+    )
+]
+
+start_time = datetime.now()
```
lucekdudek, line 35: I would move this closer to main execution loop

approxit (approved):

examples/hash_cracker_ray.py
shadeofblue: oops, good catch :sweat_smile:

PR #203, opened 2024-02-29
GPU support + examples

shadeofblue:
* add support for the `vm-nvidia` runtime
* add a GPU Dockerfile
* add examples utilizing GPU on ray-on-golem

### DIFF
--- a/examples/GPU.Dockerfile
+++ b/examples/GPU.Dockerfile
@@ -1,0 +1,43 @@
+FROM debian:bookworm-slim
+
+RUN apt-get update && apt-get install -y \
+		openssh-server \
+		iproute2 \
+		nmap \
+		tcpdump \
+		net-tools \
+		netcat-traditional \
+		screen \
+		rsyslog \
+		rsync \
+		vim \
+		curl \
+        wget \
+	&& rm -rf /var/lib/apt/lists/*
+
+RUN echo "UseDNS no" >> /etc/ssh/sshd_config && \
+	echo "PermitRootLogin yes" >> /etc/ssh/sshd_config && \
+	echo "PasswordAuthentication yes" >> /etc/ssh/sshd_config
+
+RUN wget -O miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-py310_23.11.0-2-Linux-x86_64.sh
+RUN bash miniconda.sh -b -u
+ENV PATH="/root/miniconda3/bin:${PATH}"
+RUN bash -c "echo 'PATH=/root/miniconda3/bin:${PATH}' >> /root/.bashrc"
+RUN conda install -y cudatoolkit
+
+RUN pip install wheel setuptools typing-extensions
+
+WORKDIR /app
+
+COPY pyproject.toml README.md /app/
+COPY ray_on_golem/__init__.py /app/ray_on_golem/__init__.py
+
+RUN pip install poetry && \
+	poetry config virtualenvs.create false
+RUN poetry install --no-interaction --no-ansi --only ray
+
+RUN pip install pillow
+RUN pip install numpy numba
+RUN pip config set global.index-url https://pypi.dev.golem.network/simple
+
+COPY ray_on_golem /app/ray_on_golem/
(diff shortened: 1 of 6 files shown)


examples/GPU.Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+		nmap \
+		tcpdump \
+		net-tools \
+		netcat-traditional \
+		screen \
+		rsyslog \
+		rsync \
+		vim \
+		curl \
+        wget \
```
lucekdudek, line 15: Please stick to one: tabulators or spaces

examples/gpu.yaml
```diff
@@ -0,0 +1,58 @@
...
+# The cloud provider-specific configuration properties.
+provider:
+  type: "external"
+  use_internal_ips: true
+  module: "ray_on_golem.provider.node_provider.GolemNodeProvider"
+  parameters:
+    node_config:
+      # Parameters for golem demands (same for head and workers)
+      demand:
+        image_tag: "blueshade/raytest:0.9.0-py3.10-ray2.9.2-cuda11.8"
```
lucekdudek, line 16: Can we change `raytest`?

examples/hello_gpu.py
```diff
@@ -0,0 +1,33 @@
+"""numba/cuda-on-ray equivalent of hello world"""
+import math
+
+import numpy as np
+import ray
+from numba import cuda
+
+
+def test(v: np.array):
```
lucekdudek, line 9: Rename this function or add PyDoc please

examples/hello_gpu.py
```diff
@@ -0,0 +1,33 @@
...
+    x, y = cuda.grid(2)
+    v[x, y] = 1
+
+
+@ray.remote(num_gpus=1)
+def cuda_test(width, height):
```
lucekdudek, line 15: Rename this function or add PyDoc please

examples/mandelbrot_image_gpu.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+    parser.add_argument("-r", "--use-ray", action="store_true", help="use ray")
+    parser.add_argument(
+        "-R", "--no-use-ray", dest="use_ray", action="store_false", help="don't use ray"
+    )
+    parser.add_argument("-g", "--use-gpu", action="store_true", help="use GPU")
+    parser.add_argument(
+        "-G", "--no-use-gpu", dest="use_gpu", action="store_false", help="don't use GPU"
+    )
+    parser.set_defaults(use_ray=True, use_gpu=True)
```
lucekdudek, line 248: If we are passing this argument in multiple functions maybe it would be a good idea to have one `mandelbrot_image.py` example that will cover both CPU and GPU use cases?

lucekdudek (commented):
Please see my comments

examples/mandelbrot_image_gpu.py
shadeofblue: there are two issues:

* this example requires the `numba` library which needs to be installed on top of `pillow` and obviously, requires either an image that already includes `numba` - or - requires the library to be installed when spawning the cluster
* it's inherently much more complicated that the original example...

so, consequently, I'd prefer to leave the original example in place and have this, more optimized and complicated one as an addition...

examples/gpu.yaml
shadeofblue: I'm unsure if I get the question but anyway, yeah, maybe we can upload the image to the production repo in the registry when we release a new version... otherwise, we never put images from `develop` in the `golem/ray-on-golem` repo...

lucekdudek: `raytest` is a very raw name for an image. Can we change it to something more descriptive and accurate?

examples/mandelbrot_image_gpu.py
lucekdudek: :+1: marking as resolved

lucekdudek (approved):

examples/gpu.yaml
shadeofblue: that's not the name of the image, that's the name of the repository...  but anyway, yeah, we can

examples/hello_gpu.py
shadeofblue: actually, on second thought, this one seems quite accurate...

Issue #202, opened 2024-02-29
`ray submit` on a different cluster yaml fails


shadeofblue:
Actual result:
When you try `ray submit` using a different cluster yaml file than what the cluster was spawned with, it results in an SSH password prompt

Expected result:
I'd expect the command to be somehow aware that the cluster had not been created for this particular yaml and to throw a more appropriate error to communicate the need to start this cluster first

PR #210, opened 2024-03-12
Minor fixes

### LINKED ISSUE #209: yagna payment status never ends
This issue seems to be more damaging as it is causing deadlock for long running clusters. Unfortunately, direct cause of the deadlock is unknown. 

Related subporcess:
```python
                output = json.loads(
                    await run_subprocess_output(
                        self._yagna_path,
                        "payment",
                        "status",
                        "--network",
                        network,
                        "--driver",
                        driver,
                        "--json",
                    )
                )
```
https://github.com/golemfactory/ray-on-golem/blob/main/ray_on_golem/server/services/yagna.py#L180-L192

approxit:
What I've done:
- Renamed CreateCluster to BootstrapCluster to better reflect its effect (it's not needed in the long run anyway)
- Made significantly shorter amount of time that aiohttp waits for pending connections to close while shutting down, from 1 minute to 5 seconds
- Fixed web exceptions reasons in the webserver
- Created a guard that would explicitly refuse to work if the ray is called with a different cluster name that is running on the webserver (#202)
- Added timeout for yagna status (#209)

Closes #209, #202


shadeofblue (approved):
lgtm

ray_on_golem/utils.py
```diff
@@ -27,14 +28,17 @@ async def run_subprocess(
...
         *args,
         stdout=asyncio.subprocess.PIPE,
         stderr=asyncio.subprocess.PIPE,
     )
 
-    stdout, stderr = await process.communicate()
+    stdout, stderr = await asyncio.wait_for(
+        process.communicate(),
+        timeout.total_seconds() if timeout else None,
+    )
```
lucekdudek, line 41: Shouldn't we stop the process or something like that?

lucekdudek (approved):

PR #214, opened 2024-03-14
Implement persistent, local reputation + integration with the external Reputation System

shadeofblue:
* Add Tortoise ORM + Aerich migration system
* Add FactoryBoy to facilitate easy creation of complex mock data structures
* Reputation schema, models + seed data, based on the previously-hardcoded values
* Add an adminstrative CLI tool to manage aerich migrations for the reputation db
* Add CLI to the new reputation module to manage and display the reputation data
    * add ability to update the local reputation from the remote Reputation System
* Port provider blacklist plugin to the new reputation
* Port provider scorer plugin to the new, reputation-system-based reputation
* Automatically update the reputation from the external source on webserver start
* Add factories for Demand and Proposal resources
* Add a unit tests to the provider blacklist
* Add a unit test to the reputation scorer


ray_on_golem/cli.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def with_datadir(cli_func):
+    from ray_on_golem.server.settings import DEFAULT_DATADIR
+
+    return click.option(
+        "--datadir",
+        type=Path,
+        help=f"Ray on Golem's data directory. By default, uses a system data directory: {DEFAULT_DATADIR}",
+    )(cli_func)
```
lucekdudek, line 13: I am not convinced `with_datadir` is a good name for this function.

ray_on_golem/reputation/plugins/score.py
```diff
@@ -0,0 +1,88 @@
...
+PARAM_SUCCESS_RATE = "success_rate"
+DEFAULT_UPTIME_WEIGHT: Final[float] = 1.0
+DEFAULT_SUCCESS_RATE_WEIGHT: Final[float] = 0.5
+
+
+class ReputationScorer(ProposalScorer):
+    def __init__(
+        self,
+        payment_network: str,
+        weights: Optional[Dict[str, float]] = None,
```
lucekdudek, line 25: Can you change this argument name to be more verbose. e.g. `param_weights`

ray_on_golem/reputation/plugins/score.py
```diff
@@ -0,0 +1,88 @@
...
+    ) -> float:
+        provider_id = proposal_data.issuer_id
+
+        try:
+            node_reputation = await m.NodeReputation.get_blacklisted(False).get(
+                network__network_name=self.payment_network, node__node_id=provider_id
+            )
+        except DoesNotExist:
+            node_reputation = m.NodeReputation()
+            logger.debug("Provider `%s` not found. Assuming zero scores.", provider_id)
```
lucekdudek, line 59: If scorers support `None` why not to use `None` instead of `0`?

ray_on_golem/reputation/updater.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, network: str = "polygon"):
+        self._network = network
+
+    @property
+    def reputation_uri(self):
+        return f"{REPUTATION_SYSTEM_URI}{REPUTATION_SYSTEM_PROVIDER_SCORES}?network={self._network}"
+
+    @contextmanager
+    def _no_progress_bar(self, iterable: Iterable):
+        yield iterable
```
lucekdudek, line 26: Can you add PyDoc to this method?

ray_on_golem/reputation/updater.py
```diff
@@ -0,0 +1,77 @@
...
+
+    async def update(self, progress_bar=None) -> Tuple[int, int, int, int]:
+        if not progress_bar:
+            progress_bar = self._no_progress_bar
+        cnt_added = 0
+        cnt_updated = 0
+        cnt_ignored = 0
+        cnt_total = 0
+        async with aiohttp.request("get", self.reputation_uri) as response:
```
lucekdudek, line 35: Why do we want to be in this `with` in this entire block of code? You can also create a separated method that will return `response.json()`

```python
        if not progress_bar:
            progress_bar = self._no_progress_bar
        
        data = self._get_data_from_api() # ToDo rename
        
        cnt_added = cnt_updated =  cnt_ignored = cnt_total = 0
        with progress_bar(data.get("providers")) as providers:
                        ...
```

ray_on_golem/reputation/updater.py
```diff
@@ -0,0 +1,77 @@
...
+    def _no_progress_bar(self, iterable: Iterable):
+        yield iterable
+
+    async def update(self, progress_bar=None) -> Tuple[int, int, int, int]:
+        if not progress_bar:
+            progress_bar = self._no_progress_bar
+        cnt_added = 0
+        cnt_updated = 0
+        cnt_ignored = 0
+        cnt_total = 0
```
lucekdudek, line 34: You don't need those until after you retrieve data from the API. Can you move this closer to where it is used?

Also
```python
cnt_added = cnt_updated =  cnt_ignored = cnt_total = 0
```

ray_on_golem/reputation/updater.py
```diff
@@ -0,0 +1,77 @@
...
+
+            data = await response.json()
+            with progress_bar(data.get("providers")) as providers:
+                for pdata in providers:
+                    cnt_total += 1
+                    provider_id = pdata.get("providerId")
+                    scores = pdata.get("scores")
+                    if not (provider_id and scores):
+                        cnt_ignored += 1
+                    else:
```
lucekdudek, line 49:
```suggestion
                    if not (provider_id and scores):
                        cnt_ignored += 1
                        continue
                    
                    # no else
```

lucekdudek (commented):
Most of it looks vary good, I added few comments and I am curious what you think about them.

ray_on_golem/cli.py
shadeofblue: well, that's hardly an applicable comment ;) ... and I believe the original is not bad -> it's a decorator which tells the command that it accepts the datadir option... if you don't think this is a good name, please suggest a better one, thanks :)

ray_on_golem/reputation/plugins/score.py
shadeofblue: hmm:
1. what does a score of `None` mean?
2. I believe that having no reputation _should_ be prejudiced against, as providers _should_ be interested in having the best opinion and it doesn't sound fair to promote starting from scratch e.g. to fix your reputation score...

ray_on_golem/reputation/plugins/score.py
shadeofblue: iow, to me it seems it's perfectly obvious what the weights are in the context of the reputation-based scorer...

ray_on_golem/reputation/updater.py
shadeofblue: right 😅

ray_on_golem/reputation/updater.py
shadeofblue: well, after your other suggested change, it's a one-line difference but sure ;)

shadeofblue:
@lucekdudek updated the code with the suggestions that I could :)

ray_on_golem/reputation/plugins/score.py
lucekdudek: It is not obvious that this class is based on some parameters from external system. It was not clear at all what are the keys in this dictionary. `param_weights` correspond to `key: value` of this argument while `weights` only to `values`.

ray_on_golem/reputation/plugins/score.py
lucekdudek: 1. By looking at the `ProposalScoringMixin` it looks like it is not supported anymore. @approxit do you recall what was meant to happen in that scenario?
2. Fair point

lucekdudek (approved):
Approved, please consider changing that argument name.

ray_on_golem/reputation/plugins/score.py
shadeofblue: gotcha, will update

ray_on_golem/reputation/service.py
approxit, line 1: More typing on methods would be useful.

ray_on_golem/reputation/updater.py
```diff
@@ -0,0 +1,84 @@
...
+
+import aiohttp
+
+from ray_on_golem.reputation import models as m
+
+REPUTATION_SYSTEM_URI = "https://reputation.dev-test.golem.network/v1/"
+REPUTATION_SYSTEM_PROVIDER_SCORES = "providers/scores"
+
+
+class ReputationUpdaterException(Exception):
```
approxit, line 12:
```suggestion
class ReputationUpdaterException(RayOnGolemError):
```
We are keeping all app-related exceptions on top of `RayOnGolemError`.

ray_on_golem/reputation/updater.py
```diff
@@ -0,0 +1,84 @@
...
+
+class ReputationUpdater:
+    """Simple service updating the reputation data from the remote Reputation System."""
+
+    def __init__(self, network: str = "polygon"):
+        self._network = network
+
+    @property
+    def reputation_uri(self):
+        return f"{REPUTATION_SYSTEM_URI}{REPUTATION_SYSTEM_PROVIDER_SCORES}?network={self._network}"
```
approxit, line 24: Use `yarl.URL` here instead of the raw string also, this looks like more of a static value that could be calculated once on init.

ray_on_golem/reputation/updater.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @contextmanager
+    def _no_progress_bar(self, iterable: Iterable):
+        """emulate click's progress bar interface without an actual progress bar."""
+        yield iterable
+
+    async def get_reputation_data(self):
+        """Download the reputation data from the Reputation System endpoint."""
+        async with aiohttp.request("get", self.reputation_uri) as response:
+            if not response.status == 200:
```
approxit, line 34:
```suggestion
            if response.status != 200:
```

ray_on_golem/server/main.py
```diff
@@ -158,6 +164,24 @@ async def ray_service_ctx(app: web.Application) -> None:
...
+        ReputationUpdater(network) for network in ["polygon", "mumbai", "goerli", "holesky"]
+    ]
+    update_tasks = [asyncio.create_task(updater.update()) for updater in updaters]
+
+    yield
+
+    for t in update_tasks:
+        t.cancel()
+
+    await asyncio.gather(*update_tasks)
```
approxit, line 181:
```suggestion
    await ensure_cancelled_many(update_tasks)
```
Use graceful cancel from golem-core.

ray_on_golem/server/main.py
```diff
@@ -158,6 +164,24 @@ async def ray_service_ctx(app: web.Application) -> None:
...
 
 
+async def reputation_service_ctx(app: web.Application) -> None:
+    reputation_service: ReputationService = app["reputation_service"]
+
+    await reputation_service.start()
+    updaters = [
+        ReputationUpdater(network) for network in ["polygon", "mumbai", "goerli", "holesky"]
+    ]
+    update_tasks = [asyncio.create_task(updater.update()) for updater in updaters]
```
approxit, line 174: Use `create_task_with_logging` from golem-core and also give it a proper `trace_id`.

ray_on_golem/server/services/golem/golem.py
```diff
@@ -219,19 +216,15 @@ async def _create_stack(
...
                         max_size=1000,
                         fill_at_start=True,
                         proposal_scorers=(
                             *extra_proposal_scorers.values(),
-                            MapScore(
-                                partial(
-                                    self._score_with_provider_data, payment_network=payment_network
-                                )
-                            ),
+                            ReputationScorer(payment_network=payment_network),
```
approxit, line 227:
```suggestion
                            ReputationScorer(payment_network),
```
Let's keep a consistent argument providing across plugins.

tests/unit/reputation/test_plugins_score.py
```diff
@@ -0,0 +1,67 @@
...
+            },
+            (0.0, 1.0),
+            None,
+            0.0,
+        ),
+        # zero score when not found
+        (None, None, (0.0, 1.0), DoesNotExist(), 0.0),
+    ),
+)
+@pytest.mark.asyncio
```
approxit, line 57: Its better to have
```
[tool.pytest.ini_options]
asyncio_mode = "auto"
```
in `pyproject.toml` and just use `async def` instead of explicit decorator.

approxit (commented):
I'm a fan of weights on stuff that comes from external reputation, neat!

Also, a handful of code-style and asyncio-related stuff.

ray_on_golem/reputation/plugins/score.py
```diff
@@ -0,0 +1,88 @@
...
+
+        score = (
+            sum(
+                [
+                    (getattr(node_reputation, attr) or 0.0) * weight
+                    for attr, weight in self.param_weights.items()
+                ]
```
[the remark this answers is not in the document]
shadeofblue: ah, right, that was a oversight on my part, good catch! :)

ray_on_golem/reputation/plugins/score.py
```diff
@@ -0,0 +1,88 @@
...
+            )
+            / self.total_weight
+        )
+        score_clamped = max(self.score_range[0], min(self.score_range[1], score))
```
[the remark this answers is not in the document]
shadeofblue: :woman_shrugging: it's the same for me, so why not :D

ray_on_golem/reputation/service.py
shadeofblue: I honestly think it wouldn't make much difference in case of this particular class... especially that the original methods it uses are not typed either...

ray_on_golem/reputation/updater.py
shadeofblue: I somehow don't feel that's much more readable but okay :)

ray_on_golem/server/services/golem/golem.py
shadeofblue: :woman_shrugging:

tests/unit/reputation/test_plugins_score.py
shadeofblue: I'm unsure why it's better but okay :)

shadeofblue:
@approxit updated the pull request with your suggestions

ray_on_golem/server/main.py
```diff
@@ -171,14 +171,14 @@ async def reputation_service_ctx(app: web.Application) -> None:
     updaters = [
         ReputationUpdater(network) for network in ["polygon", "mumbai", "goerli", "holesky"]
     ]
-    update_tasks = [asyncio.create_task(updater.update()) for updater in updaters]
+    update_tasks = [
+        create_task_with_logging(updater.update(), trace_id="reputation_service_ctx")
```
approxit, line 175: The trick with `trace_id` works when its values are unique. In that way, logs can differentiate task instances from each other. Refer to its usecases with `get_trace_id_name`.

ray_on_golem/reputation/updater.py
```diff
@@ -18,10 +20,11 @@ class ReputationUpdater:
 
     def __init__(self, network: str = "polygon"):
         self._network = network
-
-    @property
-    def reputation_uri(self):
-        return f"{REPUTATION_SYSTEM_URI}{REPUTATION_SYSTEM_PROVIDER_SCORES}?network={self._network}"
+        self.reputation_uri = (
+            URL(REPUTATION_SYSTEM_URI)
```
approxit, line 24: Instead of parsing `yarl.URL` here, let's just keep it as a value of constant instead - as we do for many other constants - rich object instead of raw naive value.

approxit (approved):

PR #217, opened 2024-03-25
Add ssh sentry to golem service and node monitoring to ray service

lucekdudek:


pyproject.toml
```diff
@@ -10,9 +10,9 @@ packages = [{include = "ray_on_golem"}]
 python = "^3.8.1"
 
 ray = {version = "^2.9.2", extras = ["default"]}
-#golem-core = {path = "../golem-core-python", develop = true}
-#golem-core = {git = "https://github.com/golemfactory/golem-core-python.git", branch = "main"}
-golem-core = "^0.6.0"
+# golem-core = {path = "../golem-core-python", develop = true}
+golem-core = {git = "https://github.com/golemfactory/golem-core-python.git", branch = "ld/observ-ssh-tunel"}
```
shadeofblue, line 14: please merge core first :)

pyproject.toml
```diff
@@ -22,17 +22,18 @@ dpath = "^2.1.6"
 appdirs = "^1.4.4"
 
 websocat = "^1.12.0"
-golem-node = {version = "0.15.0a18", source = "test-pypi"}
+golem-node = "0.13.2"
```
shadeofblue, line 25: I think we should not downgrade to 0.13 on develop

ray_on_golem/server/services/golem/golem.py
```diff
@@ -272,14 +280,17 @@ def _score_with_provider_data(
         return 0.5 + (0.5 * (provider_pos / len(prescored_providers)))
 
     @staticmethod
-    async def _get_provider_desc(context: WorkContext):
-        return f"{await context.get_provider_name()} ({await context.get_provider_id()})"
+    async def get_provider_desc(activity: Activity):
+        return (
+            f"{await activity.agreement.proposal.get_provider_name()}"
+            f"({await activity.agreement.proposal.get_provider_id()})"
+        )
```
shadeofblue, line 287: this would look nicer, but of course it's optional :)

```suggestion
    async def get_provider_desc(activity: Activity):
        proposal = activity.agreement.proposal
        return (f"{await proposal.get_provider_name()} ({await proposal.get_provider_id()})")
```

shadeofblue: but on the other hand, my question is why are we eliminating the call from the context?

ray_on_golem/server/services/golem/golem.py
```diff
@@ -311,14 +319,77 @@ async def _upload_node_configuration(
...
+            logger.debug(
+                f"Restarting ssh service on {provider_desc}, {ip=}, {context.activity=} done"
+            )
+        except Exception:
+            msg = f"Failed to restart SSH server {provider_desc}, {ip=}, {context.activity=}"
+            logger.warning(msg)
+            logger.debug(msg, exc_info=True)
+
+    @staticmethod
+    async def _verify_ssh_connection_check(activity_id, ssh_command):
```
shadeofblue, line 341: this function doesn't do what it says it's doing :D

maybe we should move this:
```
        ssh_command = (
            f"{get_ssh_command(ip, ssh_proxy_command, ssh_user, ssh_private_key_path)} uptime"
        )
```

inside this function and only pass the actual command there... otherwise, it's just a generic process runner which checks the process' exit code ;)

ray_on_golem/server/services/ray.py
```diff
@@ -216,6 +228,35 @@ def _get_new_node_id(self) -> NodeId:
         self._nodes_id_counter += 1
         return node_id
 
+    async def _monitor_node(self, node_id: NodeId, activity: Activity) -> None:
+        async with self._get_node_context(node_id) as node:  # type: Node
+            is_head_node = self._is_head_node(node.tags)
+        while True:
+            try:
+                activity_state = await activity.get_state()
```
shadeofblue, line 236: I wouldn't assume the activity failed until we retry this and e.g. three consecutive calls to `get_state` fail...

shadeofblue: unless of course, the function _does_ return with a state of e.g. `Terminated`  from which there's no return ;)

shadeofblue (commented):
please see my comments

ray_on_golem/server/services/ray.py
lucekdudek: get_state has 2 retries(via decorator), I can make it 3(or 5?)

ray_on_golem/server/services/ray.py
```diff
@@ -216,6 +228,35 @@ def _get_new_node_id(self) -> NodeId:
...
+            except Exception:
+                msg = (
+                    f"Activity on {await self._golem_service.get_provider_desc(activity)} "
+                    f"is no longer accessible. Terminating {node_id} {is_head_node=}"
+                )
+                logger.warning(msg)
+                logger.debug(msg, exc_info=True)
+                del self._node_monitoring_tasks[node_id]
+                if is_head_node:
+                    await self.shutdown()
```
lucekdudek, line 254: This probably needs `async with self._nodes_lock:` when I think about it. @approxit is there a better way to terminate entire cluster?

ray_on_golem/server/services/golem/golem.py
lucekdudek: > why are we eliminating the call from the context

I don't understand. I needed it to work on activity level so I decided there is no point of keeping both methods here. Context is still capable to return those values

approxit:
```suggestion
    async def get_provider_desc(activity: Activity):
        proposal = activity.agreement.proposal
        provider_name = await proposal.get_provider_name()
        provider_id = await proposal.get_provider_id()
        return f"{provider_name} ({provider_id})"
```
My "nicer" take would look like this... 😅

ray_on_golem/server/services/golem/golem.py
```diff
@@ -296,12 +307,9 @@ async def _upload_node_configuration(
         context: WorkContext,
         ip: str,
         ssh_public_key_data: str,
-        *,
-        add_state_log: Callable[[str], Awaitable[None]],
     ):
-        provider_desc = await self._get_provider_desc(context)
+        provider_desc = await self.get_provider_desc(context.activity)
         logger.info(f"Running initial commands on {provider_desc}, {ip=}, {context.activity=}")
-        await add_state_log("[6/9] Running bootstrap commands...")
```
approxit, line 304: Any reason for moving state logging outside of functions? It was aimed to keep consistency across all steps.

ray_on_golem/server/services/golem/golem.py
```diff
@@ -311,14 +319,77 @@ async def _upload_node_configuration(
...
+        logger.debug(f"Restarting ssh service on {provider_desc}, {ip=}, {context.activity=}")
+        try:
+            await context.run("service ssh restart", timeout=120)
+            logger.debug(
+                f"Restarting ssh service on {provider_desc}, {ip=}, {context.activity=} done"
+            )
+        except Exception:
+            msg = f"Failed to restart SSH server {provider_desc}, {ip=}, {context.activity=}"
+            logger.warning(msg)
+            logger.debug(msg, exc_info=True)
```
approxit, line 338:
```suggestion
        try:
            await context.run("service ssh restart", timeout=120)
        except Exception:
            msg = f"Failed to restart SSH server {provider_desc}, {ip=}, {context.activity=}"
            logger.warning(msg)
            logger.debug(msg, exc_info=True)
        else:
            logger.debug(
                f"Restarting ssh service on {provider_desc}, {ip=}, {context.activity=} done"
            )
```

ray_on_golem/server/services/golem/golem.py
approxit, line 341: The whole function is kinda implemented in `run_subprocess_output`, use it instead. Also, timeout should be provided.

ray_on_golem/server/services/golem/golem.py
```diff
@@ -311,14 +319,77 @@ async def _upload_node_configuration(
...
+            f"{get_ssh_command(ip, ssh_proxy_command, ssh_user, ssh_private_key_path)} uptime"
+        )
+
+        fails_count = 0
+        while True:
+            try:
+                await self._verify_ssh_connection_check(context.activity.id, ssh_command)
+            except Exception:
+                fails_count += 1
+                if fails_count >= 5:
```
approxit, line 375: The retry count could be extracted as a global constant in the file.

ray_on_golem/server/services/golem/golem.py
```diff
@@ -311,14 +319,77 @@ async def _upload_node_configuration(
...
+                    # Ray's autoscaler won't notice but we will at least save some GLMs
+                    msg = f"Destroying activity due to no SSH connection to {provider_desc}"
+                    logger.warning(msg)
+                    logger.debug(msg, exc_info=True)
+                    try:
+                        await context.activity.destroy()
+                    except Exception:
+                        logger.debug(f"Cannot destroy activity {provider_desc}", exc_info=True)
+                    break
```
approxit, line 384: Too many tries block could be moved out of the `while` loop for readability.

ray_on_golem/server/services/golem/golem.py
```diff
@@ -311,14 +319,77 @@ async def _upload_node_configuration(
...
+
+                logger.debug(
+                    f"SSH connection to {provider_desc} stopped working. Restarting SSH server",
+                    exc_info=True,
+                )
+                await self._restart_ssh_server(context, ip)
```
approxit, line 390: Why are we restarting the ssh server after the first failure?

It should at least try a few times to connect to rule out temporary connection issues on the network. Then, after deciding that ssh server looks like it's not working, try to restart it, and do the whole trip a few times.

Like utilizing `_verify_ssh_connection` under the hood.

ray_on_golem/server/services/ray.py
```diff
@@ -60,6 +65,8 @@ def __init__(
         self._nodes_id_counter = 0
 
         self._create_node_tasks: Dict[NodeId, asyncio.Task] = {}
+        self._node_monitoring_tasks: Dict[NodeId, asyncio.Task] = {}
+        self._node_monitoring_timeout: timedelta = DEFAULT_NODE_MONITORING_TIMEOUT
```
approxit, line 69: If this variable is not configured by arguments it is better to use global constant instead.

ray_on_golem/server/services/ray.py
```diff
@@ -196,6 +203,11 @@ async def _create_node(
             # TODO: check if node is a head node
             if not self._is_head_node_to_webserver_tunnel_running():
                 await self._start_head_node_to_webserver_tunnel()
+
+            # monitor activity state
+            self._node_monitoring_tasks[node_id] = asyncio.create_task(
```
approxit, line 208: Use `create_task_with_logging` from golem-core here. Set proper `trace_id` too.

ray_on_golem/server/services/golem/golem.py
lucekdudek: That is to much for me and the blue's suggestions fits in a single line

ray_on_golem/server/services/golem/golem.py
lucekdudek: Yes, so (almost)all are called from a single method now and it is easier to find them. A top of that it was needed to perform the little refactor around `_verify_ssh_connection_check`

ray_on_golem/server/services/golem/golem.py
lucekdudek: done

ray_on_golem/server/services/ray.py
lucekdudek: Added argument to init

ray_on_golem/server/services/ray.py
lucekdudek: add here and in other 2 palces

ray_on_golem/server/services/golem/golem.py
lucekdudek: resolved in newer commit

ray_on_golem/server/services/golem/golem.py
lucekdudek: fixed

ray_on_golem/server/services/golem/golem.py
lucekdudek: I switch `_verify_ssh_connection_check` to `_verify_ssh_connection`

ray_on_golem/server/services/ray.py
shadeofblue: hmm... I'm not fond of that pattern... it should be up to the user of the library to decide how many and if any retries should be made... of course, there can be a default but it should be overrideable from user code...

shadeofblue (approved):

ray_on_golem/server/services/golem/golem.py
```diff
@@ -55,6 +58,9 @@ def __init__(self, websocat_path: Path, registry_stats: bool):
         self._stacks_locks: DefaultDict[(str, bool), asyncio.Lock] = defaultdict(asyncio.Lock)
         self._payment_manager: Optional[PaymentManager] = None
 
+        self._ssh_sentry_tasks: Dict[str, asyncio.Task] = {}
+        self._ssh_sentry_timeout: timedelta = DEFAULT_SSH_SENTRY_TIMEOUT
```
approxit, line 62: The same comment for the different place - if this timeout is not in init, the code should use just the global constant from the file.

lucekdudek: fixed

approxit (approved):

PR #221, opened 2024-03-26
Stability fixes

approxit:
Requires https://github.com/golemfactory/golem-core-python/pull/147 to be merged.

What I've done:
- Pulled down yagna version to `0.13.3`, as allocation creation is version-sensitive
- Reworked main loop of network stats, as [#181](https://github.com/golemfactory/ray-on-golem/pull/181/files#diff-f60d4e72ce0a6202fc54ed72f75b6d4f878f99436f8797771338e8ca27a6a0bcR167-R168) exposed problem with graceful shutdown while using `asyncio.run`
- Changed flow with `_consume_draft_proposals` to use `asyncio.wait_for` instead of `asyncio.wait`, as in this way we don't need to handle its cancellation manually
- Introduced `DeviceListAllocationPayMentManager` which has allocation creation based on payment devices, instead of account lists - this change is a step towards yagna 0.15 support. Note that placement is not random, as yagna 0.15+ does not have API endpoint to fetch the device list, and it must be done at CLI level, which is used only on ray-on-golem, and not in api-centric golem-core
- Added suggested heads extra scoring to network stats
- Made proposal concurrency smaller, to not overshoot the currently available amount of suggested heads on the mainnet
- Made proposal expiration time smaller, to avoid generating warnings about too-late expiration
- Made suggested heads score boost bigger to keep them at the top of the proposal scoring
- Added `--sender` argument in case of mainnet fund preparation

Notable remarks:
- While executing `asyncio.run(func())`, `KeyboardInterrupt` can be raised in a random place (it could be somewhere in `func` or somewhere in event loop internals), stopping `func` execution. `asyncio.run` will catch that exception and will try to cancel **all remaining** tasks at once. In this way, we can't control the order of graceful shutdown. This should work in [python 3.11+](https://docs.python.org/3.12/library/asyncio-runner.html#handling-keyboard-interruption), but we are stuck in python 3.8.
- Yagna 0.15+ have reworked payment flow, where accounts are automatically prepared when needed, so `yagna payments init --sender --network polygon` is not needed at all.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,10 +9,10 @@ packages = [{include = "ray_on_golem"}]
 [tool.poetry.dependencies]
 python = "^3.8.1"
 
-ray = {version="~2.9.3", extras=["default"]}
-#golem-core = {path="../golem-core-python", develop=true}
-#golem-core = {git="https://github.com/golemfactory/golem-core-python.git", branch="main"}
-golem-core = "^0.6.2"
+ray = {version = "~2.9.3", extras = ["default"]}
+#golem-core = {path = "../golem-core-python", develop = true}
+golem-core = {git = "https://github.com/golemfactory/golem-core-python.git", branch = "main"}
+#golem-core = "^0.6.2"
 aiohttp = "^3"
 requests = "^2"
 click = "^8"
@@ -35,10 +35,10 @@ setuptools = "*"
 [tool.poetry.group.ray.dependencies]
 python = "^3.8.1"
 
-ray = {version="==2.9.3", extras=["default"]}
-#golem-core = {path="../golem-core-python", develop=true}
-#golem-core = {git="https://github.com/golemfactory/golem-core-python.git", branch="main"}
-golem-core = "^0.6.2"
+ray = {version = "==2.9.3", extras = ["default"]}
+#golem-core = {path = "../golem-core-python", develop = true}
+golem-core = {git = "https://github.com/golemfactory/golem-core-python.git", branch = "main"}
+#golem-core = "^0.6.2"
 aiohttp = "^3"
 requests = "^2"
 click = "^8"
--- a/ray_on_golem/server/services/golem/golem.py
+++ b/ray_on_golem/server/services/golem/golem.py
@@ -4,7 +4,6 @@ import logging
 from collections import defaultdict
 from datetime import datetime, timedelta, timezone
 from decimal import Decimal
-from functools import partial
 from pathlib import Path
 from typing import Awaitable, Callable, DefaultDict, Dict, Optional, Tuple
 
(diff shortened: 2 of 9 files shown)


lucekdudek (approved):

ray_on_golem/server/services/golem/golem.py
```diff
@@ -244,7 +296,7 @@ async def _create_stack(
                     ProposalBuffer(
                         min_size=0,
                         max_size=4,
-                        fill_concurrency_size=4,
+                        fill_concurrency_size=2,
```
lucekdudek, line 299: Please revert this change (as we discussed)

pyproject.toml
```diff
@@ -22,18 +22,18 @@ dpath = "^2.1.6"
 appdirs = "^1.4.4"
 
 websocat = "^1.12.0"
-golem-node = {version = "0.15.0a18", source = "test-pypi"}
+golem-node = "0.13.3"
```
shadeofblue, line 25: I'd still insist on using the newest available version of 0.15 in most of our tests, in order to have a chance to find issues with it before we publicly release a ray-on-golem version which utilizes it

ray_on_golem/server/services/golem/golem.py
```diff
@@ -46,6 +59,45 @@
...
+
+
+# FIXME: Rework this into the golem-core's DefaultPaymentManager and Allocation on yagna 0.16+,
+#  as until then there is no api call available to get driver lists and golem-core is api-only
+class DeviceListAllocationPaymentManager(DefaultPaymentManager):
+    @trace_span(show_arguments=True, show_results=True)
+    async def _create_allocation(self, budget: Decimal, network: str, driver: str) -> Allocation:
+        output = json.loads(
+            # await run_subprocess_output(YAGNA_PATH, "payment", "driver", "list", "--json")  # yagna 0.15+
+            await run_subprocess_output(YAGNA_PATH, "payment", "drivers", "--json")  # yagna 0.13
```
shadeofblue, line 73: I believe we have settled for a version check and a version dependent call here, haven't we?

ray_on_golem/server/services/golem/helpers/manager_stack.py
```diff
@@ -118,7 +118,7 @@ def _score_suggested_heads(
             proposal_data.properties.get("golem.node.debug.subnet") == priority_head_subnet_tag
         )
 
-        return 0.5 if add_scoring else 0
+        return 10 if add_scoring else 0
```
shadeofblue, line 121: please define a module-level constant to use here

ray_on_golem/server/services/golem/golem.py
```diff
@@ -255,7 +307,7 @@ async def _create_stack(
         return stack
 
     async def _get_proposal_expiration(self, proposal: Proposal) -> timedelta:
-        return await proposal.get_expiration_date() - datetime.now(timezone.utc)
+        return (await proposal.get_expiration_date() - datetime.now(timezone.utc)) * 0.8
```
shadeofblue, line 310: where does the `0.8` come from? wouldn't it make sense to define this magic number as a sensibly named constant so it's obvious what it does?

shadeofblue (commented):
please see my comments

pyproject.toml
approxit: The problem solved itself as yagna in version 0.15.0 is available publicly.

ray_on_golem/server/services/golem/golem.py
approxit: No longer needed, as yagna 0.15 is publicly available.

PR #223, opened 2024-03-29
Ssh timings

approxit:
What I've done:
- Added `ServerAliveInterval` and `ServerAliveCountMax` both for node provider's command runner and ssh commands
- Added `ClientAliveInterval` and `ClientAliveCountMax` in sshd_config to reflect similar keep alive (it is suggested to keep them lower than `Server*` counterparts).  
- Enabled integration tests on pull requests (yay!)
- Fixed problem, where YagnaService exceptions would be not caught properly at cluster bootstrap
- Changed emerging strategy in `dev_yaml` tool, to be able to remove outbound URLs for goth tests (currently outbound is not supported on goth)
- Bumped goth to 0.16.0 to support holesky and yagna in beta versions
- Uploaded new image version

Notable highlights:
- GCS / SSH commands problems are related to GolemNodeProvider being locked at `__init__`, while bootstrapping cluster and running `yagna fund` that in some cases are not working. This will be addressed shortly.

### DIFF
--- a/ray_on_golem/server/services/utils.py
+++ b/ray_on_golem/server/services/utils.py
@@ -14,6 +14,8 @@ def get_ssh_command(
             "-o StrictHostKeyChecking=no",
             "-o UserKnownHostsFile=/dev/null",
             "-o PasswordAuthentication=no",
+            "-o ServerAliveInterval=300",
+            "-o ServerAliveCountMax=5",
             f"-o {quote(proxy_command_str)}",
             f"-i {quote(str(ssh_private_key_path))}" if ssh_private_key_path else "",
             quote(ssh_user_str),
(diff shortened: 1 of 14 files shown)


lucekdudek (approved):

ray_on_golem/server/services/utils.py
```diff
@@ -14,6 +14,8 @@ def get_ssh_command(
             "-o StrictHostKeyChecking=no",
             "-o UserKnownHostsFile=/dev/null",
             "-o PasswordAuthentication=no",
+            "-o ServerAliveInterval=300",
+            "-o ServerAliveCountMax=5",
```
lucekdudek, line 18: Consider moving `300` and `5` to a config file

PR #228, opened 2024-04-04
Ray deps test

approxit:
What I've done:

- Removed all unnecessary dependencies for ray nodes
- Reworked code base to avoid dependencies requirement for ray nodes
- Added shell script to prepare and test dependencies that are used on the ray nodes 
- Added CI workflow to run the new shell script
- Removed no longer needed capabilities yaml override - as proper capabilities are prepared based on outbound URLs


shadeofblue (approved):
business-woman-ok-hand-sign-27128527

lucekdudek (approved):

PR #229, opened 2024-04-04
bump golem-node to 0.15

shadeofblue:


approxit (approved):
Note that without #221 allocations are bugged on yagna 0.15

PR #231, opened 2024-04-12
Add server warnings

lucekdudek:


ray_on_golem/server/services/ray.py
```diff
@@ -248,16 +251,25 @@ async def _monitor_node(self, node_id: NodeId, activity: Activity) -> None:
...
                 msg = (
-                    f"Activity on {await self._golem_service.get_provider_desc(activity)} "
+                    f"Activity on {provider_desc} "
                     f"is no longer accessible. Terminating {node_id} {is_head_node=}"
                 )
                 logger.warning(msg)
                 logger.debug(msg, exc_info=True)
                 if is_head_node:
+                    self.add_service_warning(
+                        f"{datetime.now(timezone.utc).isoformat()} Terminating cluster as "
```
approxit, line 263: Looks like the timestamp could be moved into `add_service_warning` as a bool with a default.

ray_on_golem/server/services/ray.py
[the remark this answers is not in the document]
lucekdudek: Done

ray_on_golem/server/services/ray.py
lucekdudek: Done

ray_on_golem/server/services/ray.py
```diff
@@ -70,23 +71,21 @@ def __init__(
...
         self._ssh_user: Optional[str] = None
+        super().__init__()
 
     async def shutdown(self) -> None:
         logger.info("Stopping RayService...")
         await self._stop_node_monitoring_tasks()
         await self._stop_head_node_to_webserver_tunnel()
         await self._stop_create_node_tasks()
         await self._destroy_nodes()
-        self._service_warnings.clear()
```
approxit: This option would be useful in the mixin.

ray_on_golem/server/services/utils.py
approxit, line 1: Would the `mixins.py` file name be more accurate in this case?

ray_on_golem/server/services/utils.py
```diff
@@ -0,0 +1,15 @@
+from datetime import datetime, timezone
+from typing import List
+
+
+class WarningMessagesMixin:
+    def __init__(self) -> None:
+        self._service_warnings = []
```
approxit, line 7:
```suggestion
        self._warning_messagess = []
```
Maybe?

approxit (approved):

ray_on_golem/server/services/utils.py
lucekdudek: Yes

ray_on_golem/server/services/utils.py
lucekdudek: Done

approxit (approved):

PR #232, opened 2024-04-16
Cluster refactor

approxit:
Warning, massive changes ahead!

What I've done:
- Made business logic object-centric instead of service-centric
- Each object is responsible for itself as much as possible
- Support for multiple clusters at the same time
- Most of the client requests are pointing to a specific cluster
- RayService handles only ray-related requests and cluster collection
- GolemService has become quite small due to logic being moved elsewhere
- Cluster is responsible for the nodes collection and manager stack collection
- ClusterNode is responsible for its state, data, activity, and sidecar tasks collection
- ClusterNodeSidecar takes responsibility for dedicated logic that is run alongside the cluster node

Notable remarks:
- Removal of all nodes in the cluster currently does nothing special - cluster will still exist and manager stacks will still run.
- Successful head node start could stop its manager stack
- New versions of the services are temporary to still have originals for reference. They will be removed before merging.


lucekdudek (commented):
Looks very good. I added few minor comments and questions.

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+    Tags,
+)
+from ray_on_golem.server.services.golem.golem import DeviceListAllocationPaymentManager
+from ray_on_golem.server.services.golem.helpers.demand_config import DemandConfigHelper
+from ray_on_golem.server.services.golem.manager_stack import ManagerStack
+from ray_on_golem.server.services.new_golem import GolemService
+
+IsHeadNode = bool
+NodeType = str
+StackKey = Tuple[NodeType, IsHeadNode]
```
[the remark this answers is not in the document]
approxit: Alias has been available since Python 3.10. Because of our golem-core's Python version agreement, I'm defaulting to 3.8 in this project too.

ray_on_golem/server/services/golem/manager_stack.py
```diff
@@ -36,3 +72,119 @@ async def stop(self) -> None:
...
+            proposal_negotiators.append(
+                MidAgreementPaymentsNegotiator(
+                    min_debit_note_interval=DEFAULT_DEBIT_NOTE_INTERVAL,
+                    optimal_debit_note_interval=DEFAULT_DEBIT_NOTE_INTERVAL,
+                    min_payment_timeout=minimal_payment_timeout,
+                    optimal_payment_timeout=optimal_payment_timeout,
+                )
+            )
+
+        demand_manager = ManagerStackNodeConfigHelper.prepare_demand_manager_for_node_type(
```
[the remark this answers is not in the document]
approxit: Woah, that's a regression since #188! Good catch! Also, it looks like managers with start/stop need some kind of warning if they are used while not running.

lucekdudek (approved):

ray_on_golem/server/services/golem/manager_stack.py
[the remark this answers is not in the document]
approxit: Whoops made a change too quickly and did not see that coming! Demand manager is already added to the stack... but inside of the helper, due to conditionally having multiple managers inside.

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+        self._golem_service = golem_service
+        self._manager_stack = manager_stack
+
+        self._sidecars = self._prepare_sidecars()
+
+        with self.ssh_public_key_path.open() as f:
+            self._ssh_public_key_data = f.readline().strip()
+
+    def __str__(self) -> str:
+        return self.node_id
```
shadeofblue, line 68:
```suggestion
        return f"{self.__class__.__name__}: {self.node_id}"
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+    def get_warning_messages(self) -> List[str]:
+        warnings = super().get_warning_messages()
+
+        for sidecar in self._sidecars:
+            warnings.extend(sidecar.get_warning_messages())
+
+        return warnings
+
+    def schedule_start(self) -> None:
+        if self._start_task and not self._start_task.done():
```
shadeofblue, line 86: doesn't that mean that if `_start_task` is done, it will be created again? is that intended?

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+            logger.info(f"Not scheduling start `%s` node, as it's already scheduled", self)
```
shadeofblue, line 87:
```suggestion
            logger.info(f"Node `%s`: start already scheduled", self)
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+                    logger.debug(msg, exc_info=True)
+                else:
+                    break
+        except Exception as e:
+            self.state = NodeState.terminated
+            self._add_state_log(
+                f"Failed to create activity: {type(e).__module__}.{type(e).__name__}: {e}"
+            )
+
+            logger.info("Starting `%s` node failed!", self)
```
shadeofblue, line 125:
```suggestion
            logger.info("Node `%s` startup failed!", self)
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+            self.start(),
+            trace_id=get_trace_id_name(self, "scheduled-start"),
+        )
+
+    async def start(self) -> None:
+        if self.state in (NodeState.running, NodeState.terminating):
+            logger.info(f"Not starting `%s` node, as it's already running or stopping", self)
+            return
+
+        logger.info("Starting `%s` node...", self)
```
shadeofblue, line 100:
```suggestion
        logger.info("Starting node `%s`...", self)
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+
+        self._activity = activity
+
+        await self._start_sidecars()
+
+        logger.info("Starting `%s` node done", self)
+
+    async def stop(self) -> None:
+        if self.state in (NodeState.terminating, NodeState.terminated):
+            logger.info(f"Not stopping `%s` node, as it's already stopped", self)
```
shadeofblue, line 143: same here, could be simplified to: 

```suggestion
            logger.info(f"Node `%s` already terminated.", self)
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+            return
+
+        logger.info("Stopping `%s` node...", self)
```
shadeofblue, line 146:
```suggestion
        logger.info("Stopping node `%s`...", self)
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+
+        if self._activity:
+            await self._stop_activity(self._activity)
+            self._activity = None
+
+        self.state = NodeState.terminated
+        self.internal_ip = None
+        self.ssh_proxy_command = None
+
+        logger.info("Stopping `%s` node done", self)
```
shadeofblue, line 164:
```suggestion
        logger.info("Node `%s` stopped.", self)
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,377 @@
...
+
+    def _add_state_log(self, log_entry: str) -> None:
+        self.state_log.append(log_entry)
+
+    @staticmethod
+    async def get_provider_desc(activity: Activity) -> str:
+        proposal = activity.agreement.proposal
+        return f"{await proposal.get_provider_name()} ({await proposal.get_provider_id()})"
```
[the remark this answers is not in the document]
shadeofblue: activity? or `Proposal`?

ray_on_golem/server/cluster/nodes.py
```diff
@@ -0,0 +1,381 @@
...
+
+        hostname = ip.replace(".", "-")
+
+        await context.run("echo 'ON_GOLEM_NETWORK=1' >> /etc/environment")
+        await context.run(f"echo 'NODE_IP={ip}' >> /etc/environment")
+        await context.run(f"hostname '{hostname}'")
+        await context.run(f"echo '{hostname}' > /etc/hostname")
+        await context.run(f"echo '{ip} {hostname}' >> /etc/hosts")
+        await context.run("mkdir -p /root/.ssh")
+        await context.run(f'echo "{ssh_public_key_data}" >> /root/.ssh/authorized_keys')
```
shadeofblue, line 254: this code, and possibly other code in this module, is no longer valid...

please have a look at the current version: https://github.com/golemfactory/ray-on-golem/blob/main/ray_on_golem/server/services/golem/golem.py#L336

ray_on_golem/server/cluster/sidecars.py
```diff
@@ -0,0 +1,249 @@
...
+
+        self._monitor_task: Optional[asyncio.Task] = None
+
+    @abstractmethod
+    async def _monitor(self) -> None:
+        ...
+
+    async def start(self) -> None:
+        if self.is_running():
+            logger.info(f"Not starting `%s` node {self.name}, as it's already running", self._node)
```
shadeofblue, line 53: is this about starting the _node_? or the node's specific monitor?

ray_on_golem/server/cluster/sidecars.py
```diff
@@ -0,0 +1,249 @@
...
+                )
+
+                provider_desc = await self._node.get_provider_desc(self._node.activity)
+
+                if isinstance(self._node, HeadClusterNode):
+                    self.add_warning_message(
+                        f"Terminating whole cluster as Head({self._node.node_id}) {provider_desc}"
+                        " activity is no longer accessible"
+                    )
+                    create_task_with_logging(self._cluster.stop())
```
shadeofblue, line 113: it seems wrong to me for the node's monitor to signal the cluster to stop... 

this reversal of control seems very hard to maintain in the long run -> if we add such triggers in multiple monitors, it will be cumbersome to change the behavior in case of wider changes... 

rather the other way round, the cluster should detect that its head node is corrupted and should decide to stop. The logic triggering cluster abort should ideally be gathered in one place.

ray_on_golem/server/services/new_golem.py
```diff
@@ -0,0 +1,117 @@
...
+EXPIRATION_TIME_FACTOR = 0.8
+
+
+class NoMatchingPlatform(AllocationException):
+    ...
+
+
+# FIXME: Rework this into the golem-core's DefaultPaymentManager and Allocation on yagna 0.16+,
+#  as until then there is no api call available to get driver lists and golem-core is api-only
+class DeviceListAllocationPaymentManager(DefaultPaymentManager):
```
shadeofblue, line 38: what's "DeviceList" here? please add a docstring to this class...

shadeofblue (commented):
added comments, please also add docstrings to classes and methods if possible

shadeofblue (changes requested):
please remove the no-longer-used code and please merge newest changes from develop and make sure the code here reflects those changes :/

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+        warnings = super().get_warning_messages()
+
+        for node in self._nodes.values():
+            warnings.extend(node.get_warning_messages())
+
+        return warnings
+
+    async def start(self) -> None:
+        if self._state in (NodeState.pending, NodeState.running):
+            logger.info("Not starting `%s` cluster as it's already running or starting", self._name)
```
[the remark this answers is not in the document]
approxit: This form is used across the project. I'm not trying to introduce a new pattern.

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+
+        await self._payment_manager.start()
+
+        self._state = NodeState.running
+
+        logger.info("Starting `%s` cluster done", self._name)
+
+    async def stop(self) -> None:
+        if self._state in (NodeState.terminating, NodeState.terminated):
+            logger.info("Not stopping `%s` cluster as it's already stopped or stopping", self._name)
```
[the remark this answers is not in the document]
approxit: I'm afraid that I need to disagree. Here, the log is related to the actual condition. Your proposal can be misleading in case the cluster was never started.

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+        for node in self._nodes.values():
+            await node.stop()
+
+        self._nodes.clear()
+        self._nodes_id_counter = 0
+
+        for stack in self._manager_stacks.values():
+            await stack.stop()
+
+        self._manager_stacks.clear()
```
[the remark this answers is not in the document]
approxit: I'm afraid that I need to disagree here too. I want to explicitly tear down everything and then handle stuff that needs to be paid. If we change the order, we could and up with stuff that is not paid.

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+        return (
+            partial(
+                HeadClusterNode,
+                webserver_port=self._webserver_port,
+            )
+            if is_head_node
+            else WorkerClusterNode
+        )
+
+    def _is_head_node(self, tags: Tags) -> bool:
```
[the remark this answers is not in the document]
approxit: Separate function then.

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+    NodeState,
+    ProviderParametersData,
+    Tags,
+)
+from ray_on_golem.server.services.golem.golem import DeviceListAllocationPaymentManager
+from ray_on_golem.server.services.golem.helpers.demand_config import DemandConfigHelper
+from ray_on_golem.server.services.golem.manager_stack import ManagerStack
+from ray_on_golem.server.services.new_golem import GolemService
+
+IsHeadNode = bool
```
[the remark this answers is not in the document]
approxit: I'm using the type alias to help readability. Without that, the role of `bool` needs to be read from the context, which I found problematic a few times in this area.

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+                tags=tags,
+                node_config=node_config,
+                ssh_private_key_path=self._provider_parameters.ssh_private_key,
+                ssh_public_key_path=self._provider_parameters.ssh_private_key.with_suffix(".pub"),
+                ssh_user=self._provider_parameters.ssh_user,
+            )
+
+            self._nodes[node_id] = node
+
+            node.schedule_start()
```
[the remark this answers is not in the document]
approxit: Here, the code is synchronous and there is no difference between the two versions.

ray_on_golem/server/cluster/nodes.py
approxit: The `__str__` is aimed to be as simple as possible, and it is not used as `__repr__`.

ray_on_golem/server/cluster/sidecars.py
approxit: The `self.name` holds the name of a monitor.

ray_on_golem/server/services/new_golem.py
approxit: It should be `DriverList`.

ray_on_golem/server/cluster/sidecars.py
approxit: That is fair. It should be done at the very beginning.

golem-cluster.full.yaml
```diff
@@ -1,6 +1,6 @@
 # This is Ray on Golem Full yaml 
 # - the example config for testnet cluster, 
-# - with all the properties potentially changable by the user
+# - with all the properties potentially changeable by the user
 
 # Ray on Golem cluster name
 cluster_name: golem-cluster
```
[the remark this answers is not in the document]
shadeofblue: @approxit you haven't updated the name of the cluster...

ray_on_golem/server/cluster/cluster.py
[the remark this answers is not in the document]
shadeofblue: in my opinion, this _doesn't_ improve readability... it _worsens_ it :(

normally, when I read: 

```
    async def _get_or_create_manager_stack(
...
        is_head_node: bool
    ) -> ManagerStack:
```

i immediately know what `is_head_node` argument means and what values are passed... 

with your proposed version, it's

```
    async def _get_or_create_manager_stack(
...
        is_head_node: IsHeadNode,
    ) -> ManagerStack:
```

and I have no idea what `IsHeadNode` is ... which necessitates my distracting myself with having to check it... I strongly believe it's a very bad use for a custom type...

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+            warnings.extend(node.get_warning_messages())
+
+        return warnings
+
+    async def start(self) -> None:
+        if self._state in (NodeState.pending, NodeState.running):
+            logger.info("Not starting `%s` cluster as it's already running or starting", self._name)
+            return
+
+        logger.info("Starting `%s` cluster...", self._name)
```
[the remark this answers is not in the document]
shadeofblue: same here - it's not really a natural and gramatically correct form to say "Starting X cluster" if "X" is the name of the cluster...

ray_on_golem/server/cluster/cluster.py
```diff
@@ -0,0 +1,228 @@
...
+
+        self._state = NodeState.pending
+
+        await self._payment_manager.start()
+
+        self._state = NodeState.running
+
+        logger.info("Starting `%s` cluster done", self._name)
```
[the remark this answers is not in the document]
shadeofblue: again, can we please start improving our logs _while_ we introduce changes to the code?

ray_on_golem/server/cluster/cluster.py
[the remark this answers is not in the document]
shadeofblue: well, the message above _also_ implies that the cluster has been started at some point...

shadeofblue (approved):
for me, the forms used in the logs just don't make sense... in my opinion, the ability to sieve through and group logs should not be achieved by making the logs sound awkward and un-gramatical ... so if that's the way we want them to be, I don't feel it makes sense for me try to review them...

PR #233, opened 2024-04-17
fix container initialization to include the update `pip` config

shadeofblue:


lucekdudek (approved):

approxit:
I don't see why we must be selective regarding `/root` contents. We already removed `/root/.cache` from its contents at the Dockerfile stage as we know we don't want that, but the rest? I think it should go as is.

shadeofblue:
@approxit for some reason, non-selective command fails and it's not immediately obvious why...

PR #238, opened 2024-04-18
GCS port exposure

approxit:
What I've done:
- Added option to expose GCS ports to the local (requestor) machine.

Notable remarks:
- `ray status --address=localhost:6379` can check cluster state by using the exposed GCS port (which has binary payloads)
- `ray job submit --address http://localhost:8265 --working-dir . -- python calculate.py` can submit the job by using the `ray dashboard` command (which has http payloads)

### DIFF
--- a/ray_on_golem/server/cluster/nodes.py
+++ b/ray_on_golem/server/cluster/nodes.py
@@ -463,14 +463,14 @@ class HeadClusterNode(WorkerClusterNode):
             PortTunnelClusterNodeSidecar(node=self, local_port=self._webserver_port, reverse=True),
         ]
 
-        if self._ray_gcs_expose_port is not None:
+        if self._ray_gcs_expose_port:
             sidecars.append(
                 PortTunnelClusterNodeSidecar(
                     node=self, local_port=self._ray_gcs_expose_port, remote_port=RAY_GCS_PORT
                 ),
             )
 
-        if self._ray_dashboard_expose_port is not None:
+        if self._ray_dashboard_expose_port:
             sidecars.append(
                 PortTunnelClusterNodeSidecar(
                     node=self,
--- a/ray_on_golem/server/cluster/nodes.py
+++ b/ray_on_golem/server/cluster/nodes.py
@@ -14,8 +14,8 @@ from ray_on_golem.exceptions import RayOnGolemError
 from ray_on_golem.server.cluster.sidecars import (
     ActivityStateMonitorClusterNodeSidecar,
     ClusterNodeSidecar,
-    HeadNodeToWebserverTunnelClusterNodeSidecar,
     MonitorClusterNodeSidecar,
+    PortTunnelClusterNodeSidecar,
     SshStateMonitorClusterNodeSidecar,
 )
 from ray_on_golem.server.mixins import WarningMessagesMixin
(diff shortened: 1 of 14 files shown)


ray_on_golem/server/cluster/nodes.py
```diff
@@ -435,13 +439,47 @@ async def _on_monitor_check_failed(
 class HeadClusterNode(WorkerClusterNode):
     """Self-contained element that represents explicitly a Ray head node."""
 
-    webserver_port: int
+    _webserver_port: int
```
shadeofblue, line 442: why not leave these attributes as regular ones, without making them "private"?

ray_on_golem/server/cluster/sidecars.py
```diff
@@ -182,16 +185,25 @@ async def _monitor(self) -> None:
...
+        local_port: int,
+        remote_port: Optional[int] = None,
+        reverse: bool = False,
+        **kwargs,
+    ) -> None:
+        super().__init__(**kwargs)
 
-        self._head_node = head_node
+        self._local_port = local_port
+        self._remote_port = local_port if remote_port is None else remote_port
```
shadeofblue, line 205:
```suggestion
        self._remote_port = remote_port or local_port
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -435,13 +439,47 @@ async def _on_monitor_check_failed(
...
     def _prepare_sidecars(self) -> Collection[ClusterNodeSidecar]:
-        return (
+        sidecars = [
             *super()._prepare_sidecars(),
-            HeadNodeToWebserverTunnelClusterNodeSidecar(self),
-        )
+            PortTunnelClusterNodeSidecar(node=self, local_port=self._webserver_port, reverse=True),
+        ]
+
+        if self._ray_gcs_expose_port is not None:
```
shadeofblue, line 466: why `is not None`? you can't open port 0 anyway ... 

```suggestion
        if self._ray_gcs_expose_port:
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -435,13 +439,47 @@ async def _on_monitor_check_failed(
...
+            sidecars.append(
+                PortTunnelClusterNodeSidecar(
+                    node=self, local_port=self._ray_gcs_expose_port, remote_port=RAY_GCS_PORT
+                ),
+            )
+
+        if self._ray_dashboard_expose_port is not None:
```
shadeofblue, line 473:
```suggestion
        if self._ray_dashboard_expose_port:
```

ray_on_golem/server/services/__init__.py
```diff
@@ -13,6 +13,7 @@
     "DriverListAllocationPaymentManager",
     "GolemService",
     "ManagerStack",
+    "ManagerStackNodeConfigHelper",
```
shadeofblue, line 16: what's this?

ray_on_golem/server/cluster/nodes.py
approxit: Previously that port was used externally, but not anymore. Now it exists only for the sake of sidecar, also making it private does not generate additional data in communication via http.

ray_on_golem/server/services/__init__.py
approxit: Helper class used in the netstat.

shadeofblue (approved):

PR #241, opened 2024-05-07
Cluster cleanup

approxit:
What I've done:
- Added cluster or manager stack teardown on node stop
- Reintroduced hash-based manager stack to support different versions of stacks under the same name
- Fixed mistyped condition in `cluster.clean()`
- Removed excessive dashboard exposure in favor of `ray dashboard` usage
- Normalized usage of the string quotes and bool casing in yaml files

### DIFF
--- a/ray_on_golem/server/cluster/cluster.py
+++ b/ray_on_golem/server/cluster/cluster.py
@@ -274,7 +274,7 @@ class Cluster(WarningMessagesMixin):
     def _get_hash_from_node_config(node_config: NodeConfigData) -> str:
         return hashlib.md5(node_config.json().encode()).hexdigest()
 
-    async def _on_node_stop(self, node: ClusterNode) -> None:
+    def _on_node_stop(self, node: ClusterNode) -> None:
         non_terminated_nodes = self.get_non_terminated_nodes()
         if not non_terminated_nodes:
             logger.debug("No more nodes running on the cluster, scheduling cluster stop")
@@ -283,7 +283,9 @@ class Cluster(WarningMessagesMixin):
                 self.stop(),
                 trace_id=get_trace_id_name(self, "on-node-stop-cluster-stop"),
             )
-        elif not any(node.manager_stack == n.manager_stack for n in self.get_non_terminated_nodes()):
+        elif not any(
+            node.manager_stack == n.manager_stack for n in self.get_non_terminated_nodes()
+        ):
             logger.debug(
                 "No more nodes running on the manager stack, scheduling manager stack stop"
             )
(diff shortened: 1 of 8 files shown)


ray_on_golem/server/cluster/cluster.py
```diff
@@ -243,3 +266,33 @@ async def _remove_manager_stack(self, stack_key: StackKey) -> None:
...
         logger.info(f"Removing stack `%s` done", stack_key)
+
+    def _get_stack_key(self, node_config: NodeConfigData, is_head_node: IsHeadNode) -> StackKey:
+        return (self._get_hash_from_node_config(node_config), is_head_node)
+
+    @staticmethod
+    def _get_hash_from_node_config(node_config: NodeConfigData) -> str:
+        return hashlib.md5(node_config.json().encode()).hexdigest()
+
+    async def _on_node_stop(self, node: ClusterNode) -> None:
```
lucekdudek, line 277: Multiple calls to this function should always be serial. Does it have to be async?

approxit: Well, there is no need for async, indeed. But as a sync method, it can't be called concurrently, so no locks are required.

ray_on_golem/server/cluster/cluster.py
```diff
@@ -17,15 +20,15 @@
...
     Tags,
 )
-from ray_on_golem.server.services.golem import (
+from ray_on_golem.server.services import (
     DriverListAllocationPaymentManager,
     GolemService,
     ManagerStack,
 )
 
 IsHeadNode = bool
```
shadeofblue, line 29: I still believe it makes absolutely zero sense to assign `bool` to another, custom type name...

approxit: This is changed in #242. I'll keep the current state here to avoid merge conflicts.

ray_on_golem/server/cluster/cluster.py
```diff
@@ -243,3 +266,37 @@ async def _remove_manager_stack(self, stack_key: StackKey) -> None:
             del self._manager_stacks_locks[stack_key]
 
         logger.info(f"Removing stack `%s` done", stack_key)
+
+    def _get_stack_key(self, node_config: NodeConfigData, is_head_node: IsHeadNode) -> StackKey:
```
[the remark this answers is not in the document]
approxit: This is changed in https://github.com/golemfactory/ray-on-golem/pull/242. I'll keep the current state here to avoid merge conflicts.

shadeofblue (approved):

PR #242, opened 2024-05-13
Priority subnet tag

approxit:
What I've done:
- Acquiring priority head node will now be done in an explicitly separate step instead of demand aggregation result
- Using priority subnet is now a generic setup of any node type, where the head node is the default
- Getting agreement for priority node is limited by time, default is 30 seconds

### DIFF
--- a/ray_on_golem/provider/node_provider.py
+++ b/ray_on_golem/provider/node_provider.py
@@ -46,9 +46,14 @@ PROVIDER_DEFAULTS = {
     "payment_driver": PAYMENT_DRIVER_ERC20,
     "node_config": {
         "subnet_tag": "public",
-        "priority_head_subnet_tag": "ray-on-golem-heads",
     },
     "total_budget": 1.0,
+}
+
+HEAD_NODE_DEFAULTS = {
+    "node_config": {
+        "priority_subnet_tag": "ray-on-golem-heads",
+    }
 }
 
 
@@ -290,6 +295,13 @@ class GolemNodeProvider(NodeProvider):
 
         for node_type in config.get("available_node_types", {}).values():
             node_config = deepcopy(config["provider"]["parameters"]["node_config"])
+
+            if node_type == "ray.head.default":
+                dpath.merge(
+                    node_config,
+                    deepcopy(HEAD_NODE_DEFAULTS)
+                )
+
             dpath.merge(
                 node_config,
                 node_type["node_config"],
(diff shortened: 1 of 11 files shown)


ray_on_golem/provider/node_provider.py
```diff
@@ -290,6 +295,10 @@ def _apply_config_defaults(config: Dict[str, Any]) -> None:
 
         for node_type in config.get("available_node_types", {}).values():
             node_config = deepcopy(config["provider"]["parameters"]["node_config"])
+
+            if node_type == "ray.head.default":
+                dpath.merge(node_config, deepcopy(HEAD_NODE_DEFAULTS))
```
shadeofblue, line 300: @approxit doesn't this mean that the setting defined in `HEAD_NODE_DEFAULTS` will _always_ override whatever had been set in the node config in YAML?

ray_on_golem/server/cluster/cluster.py
```diff
@@ -110,11 +109,17 @@ async def stop(self, clear: bool = True) -> None:
 
         self._state = NodeState.terminating
 
+        stop_coros = []
         for node in self._nodes.values():
-            await node.stop()
+            stop_coros.append(node.stop(call_events=False))
 
+        await asyncio.gather(*stop_coros)
```
shadeofblue, line 116: if we're only building the list to then call `asyncio.gather` with it, maybe it doesn't make sense to keep the intermediary list and just call the code with a list comprehension?
```suggestion
        await asyncio.gather(*[node.stop(call_events=False) for node in self._nodes.values()])
```

ray_on_golem/server/cluster/nodes.py
```diff
@@ -256,6 +266,25 @@ async def _create_activity(self) -> Tuple[Activity, str, str]:
 
         return activity, ip, ssh_proxy_command
 
+    async def _get_agreement(self) -> Agreement:
+        if self._priority_manager_stack:
+            self._add_state_log("[1/9] Getting agreement from providers in priority subnet...")
```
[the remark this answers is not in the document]
shadeofblue: but honesly, I would rather stay with "Getting agreement..." because, from the user's perspective, this is just this step where our aim is to acquire an agreement with a provider who supplies the head node -> we can explain in the error message after the timeout that we've failed to sign an agreement with the providers from the priority subnet and that we're retrying the step with the regular nodes... but otherwise, having this content of the first step is too much information for a regular user...

shadeofblue (approved):
looks okay, please consider my comments though

PR #243, opened 2024-05-17
Pip volumes fix

approxit:
This PR addresses the problem of the inability to install any Python packages on the running node.

What I've done:
- Added virtual env with `--system-site-packages` option to the image
- Added auto venv activation in `.bashrc` (used by ray, as it uses ssh everywhere)
- Added `/root/.local/bin` in `PATH` to stop pip complaining about not being in the path.
- Removed leftover debug params from #238

Notable remarks:
- Keeping venv on golem volume has some speed penalty, but it's an acceptable workaround until golem receives more rich functionality about volumes and file system on the runtime.

### DIFF
--- a/ray_on_golem/server/services/golem/golem.py
+++ b/ray_on_golem/server/services/golem/golem.py
@@ -104,4 +104,4 @@ class GolemService:
 
     def get_ssh_proxy_command(self, connection_uri: URL) -> str:
         # Using single quotes for the authentication token as double quotes are causing problems with CLI character escaping in ray
-        return f"{self._websocat_path} asyncstdio: {connection_uri}/22 --binary -H=Authorization:'Bearer {self._yagna_appkey}'"
+        return f"{self._websocat_path} -vv asyncstdio: {connection_uri}/22 --binary -H=Authorization:'Bearer {self._yagna_appkey}'"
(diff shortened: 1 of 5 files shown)


lucekdudek (approved):

ray_on_golem/server/services/golem/golem.py
```diff
@@ -104,4 +104,4 @@ def get_connection_uri(self, ip: str) -> URL:
 
     def get_ssh_proxy_command(self, connection_uri: URL) -> str:
         # Using single quotes for the authentication token as double quotes are causing problems with CLI character escaping in ray
-        return f"{self._websocat_path} -vv asyncstdio: {connection_uri}/22 --binary -H=Authorization:'Bearer {self._yagna_appkey}'"
+        return f"{self._websocat_path} asyncstdio: {connection_uri}/22 --binary -H=Authorization:'Bearer {self._yagna_appkey}'"
```
shadeofblue, line 107: @approxit please don't remove that flag, it makes debugging easier in case of issues

ray_on_golem/utils.py
```diff
@@ -80,7 +80,6 @@ def get_ssh_command_args(
 ) -> List[str]:
     return [
         "ssh",
-        "-vvv",
```
shadeofblue, line 83: @approxit and same here

PR #246, opened 2024-06-04
Cluster cleanup

approxit:
What I've done:
- Made that empty `Cluster` instead of being kept in the memory, it could be auto-removed after a specific time or be overridden by a new cluster before that timeout. This helps with consecutive `ray down` and `ray up`.
- Reworked `on_stop` events in `Cluster`, `ClusterNode`, and `ClusterNodeSidecar. They spawn separate asyncio tasks by default.
- Moved monitor check failed logic from particular sidecar to its parent to better separation of concerns.
- Fixed default head node `priority_subnet_tag` not being correctly handled.
- Moved site packages installed from the system level to be installed in virtualenv that is copied over to golem volume. It can take around 2 minutes for some providers, but it is necessary to have pip working while we wait for more functionality in golem volumes.
- Bumped `total_budget` from `1` to `5`.
- Removed some leftovers from workflows

Notable remarks:
- Consecutive `ray up` with the same name but with different configs results in ray-specific errors, we decided that currently, we will not implement any additional warnings/guards about that.


golem-cluster.yaml
```diff
@@ -25,7 +25,7 @@ provider:
     #payment_network: "polygon"
 
     # Maximum amount of GLMs that's going to be spent for the whole cluster
-    total_budget: 1
+    total_budget: 5
```
lucekdudek, line 28: What about `mini` and `full` configs?

lucekdudek (approved):

golem-cluster.yaml
mateuszsrebrny: in `mini` there is no total_budget (default is used), `full` already has `5` :)

ray_on_golem/server/services/ray.py
```diff
@@ -209,3 +227,27 @@ async def _with_cluster_node_context(
...
+            del self._clusters[cluster_name]
+
+        # cleanup the lock after removing the cluster
+        if not self._clusters_locks[cluster_name].locked():
+            del self._clusters_locks[cluster_name]
+
+    async def _on_cluster_stop(self, cluster: Cluster) -> None:
+        cluster_name = str(cluster)
+
+        logger.info(f"Cluster `%s` stopped, it will be removed after 30 seconds", cluster_name)
```
shadeofblue, line 247: please use the defined timeout constant instead of hardcoding the text

shadeofblue (approved):
lgtm. please apply my comment, though

PR #248, opened 2024-06-17
Less verbose initial commands output

lucekdudek:


shadeofblue (approved):
seconding @approxit 's question, though ;)

PR #257, opened 2024-08-14
fix `DriverListAllocationPaymentManager._get_allocation`

shadeofblue:
fix for: `      node0: Failed to create activity: builtins.TypeError: DriverListAllocationPaymentManager._create_allocation() missing 3 required positional arguments: 'budget', 'network', and 'driver'` 

found while testing the 0.16 rc

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Start Goth
         env:
-          GOTH_VERSION: 0.16.0
+          GOTH_VERSION: 0.19.0
           GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_ACTOR: ${{ github.actor }}
         run: ./.github/workflows/start-goth.sh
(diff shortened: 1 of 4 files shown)


lucekdudek (approved):
LGTM but @approxit had some specific idea for this class in regard to Yagna versions so please wait for his review 😄

approxit (changes requested):
The problem is not related to the yagna version, it is related to the older version of golem-core. golem-core `0.7.5` and [`DriverListAllocationPaymentManager` from golem-workers](https://github.com/golemfactory/golem-workers/blob/27d162454e777c435b96a09b2b8b09badb24ccca/golem_workers/golem.py#L37) should do the job.

shadeofblue:
> The problem is not related to the yagna version, it is related to the older version of golem-core. golem-core `0.7.5` and [`DriverListAllocationPaymentManager` from golem-workers](https://github.com/golemfactory/golem-workers/blob/27d162454e777c435b96a09b2b8b09badb24ccca/golem_workers/golem.py#L37) should do the job.

well, given that you have not moved the class to `ray-on-golem`, I believe the fix is as good as mine

